PR #17, opened 2019-02-15
Add support for encrypted backups

stevesoltys:
1. Add prompt for entering password during backup and restore.
2. Use PBKDF2 to generate a secret key that is used to encrypt backups.
3. Store salt in backup zip file.
4. Fetch salt from backup zip file during restore and use it to decrypt restoration data.


dan-v:
This is awesome @stevesoltys! Finally got around to putting your backup utility into my builds recently and it works great and encrypted backups would make this perfect.

stevesoltys:
Good to hear @dan-v. I think I want a bit more cleanup before merging this. I'll get around to it sometime this week.

Until then you could do a build on this branch if you wanted, encryption is working well.

PR #25, opened 2019-06-05
Store backup passphrase insecurely for now

grote:
This is being done to implement automatic background updates
and not supposed to be part of a release.

The backup key will later be generated and shown to the user instead of
allowing them to choose their own.

### DIFF
--- a/app/src/main/java/com/stevesoltys/backup/activity/backup/CreateBackupActivityController.java
+++ b/app/src/main/java/com/stevesoltys/backup/activity/backup/CreateBackupActivityController.java
@@ -7,12 +7,19 @@ import android.os.RemoteException;
 import android.text.InputType;
 import android.util.Log;
 import android.view.View;
-import android.widget.*;
+import android.widget.ArrayAdapter;
+import android.widget.EditText;
+import android.widget.ListView;
+import android.widget.PopupWindow;
+import android.widget.TextView;
+import android.widget.Toast;
+
 import com.google.android.collect.Sets;
 import com.stevesoltys.backup.R;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/stevesoltys/backup/activity/backup/CreateBackupActivityController.java
```diff
@@ -77,9 +93,16 @@ void showEnterPasswordAlert(Set<String> selectedPackages, Uri contentUri, Activi
                 .setMessage("You'll need this to restore your backup, so write it down!")
                 .setView(passwordTextView)
 
-                .setPositiveButton("Set password", (dialog, button) ->
+                .setPositiveButton("Set password", (dialog, button) -> {
+                    if (passwordTextView.getText().length() == 0) {
+                        Toast.makeText(parent, "Please enter a password", Toast.LENGTH_SHORT).show();
+                        dialog.cancel();
+                        showEnterPasswordAlert(selectedPackages, contentUri, parent);
```
grote: I didn't find how an empty password would disable encryption. It would just crash, so for now I added this little part forcing the user to enter a password.

stevesoltys: It seems it was broken at some point recently (probably by me, honestly) but it used to only encrypt the data if the secret key was available: https://github.com/stevesoltys/backup/blob/master/app/src/main/java/com/stevesoltys/backup/transport/component/provider/ContentProviderBackupComponent.java#L213

PR #27, opened 2019-06-12
Background backups

### LINKED ISSUE #15: Add support for targeting latest SDK version
Currently, the application crashes when the target SDK version is set to anything greater than 26.

grote:
This MR adds a button for doing daily background updates using Android's JobScheduler. The button itself will go away as background updates will become the default. It just exists for testing purposes.

The foreground service is removed as it would require to show a notification in the status bar when targeting the latest SDK API. For now, the service is started before it is used to ensure that it is running. If this will show issues, an alternative is to add a configuration that allows the app to run in the background.

Thus, the target API is changed to 28. This closes #15 

Finally, to avoid issues with backup transport re-initialization and non-functional stub components being used during the life of the transport, its components have been refactored to be an immutable part of the transport.


app/src/main/java/com/stevesoltys/backup/transport/component/provider/ContentProviderRestoreComponent.java
```diff
@@ -64,14 +81,16 @@ public int startRestore(long token, PackageInfo[] packages) {
...
         restoreState.setPackageIndex(-1);
 
-        if (configuration.getPassword() != null && !configuration.getPassword().isEmpty()) {
+        String password = requireNonNull(this.password);
+
+        if (!password.isEmpty()) {
             try {
                 ParcelFileDescriptor inputFileDescriptor = buildInputFileDescriptor();
                 ZipInputStream inputStream = buildInputStream(inputFileDescriptor);
                 seekToEntry(inputStream, ContentProviderBackupConstants.SALT_FILE_PATH);
```
stevesoltys: Could either remove `ContentProviderBackupConstants`, or migrate over any constants you think should be there. (your preference - either is fine for me)

stevesoltys: Unless this is the only shared constant between backups / restores? That would actually make sense for `ContentProviderBackupConstants`.

stevesoltys:
Pretty much entirely nitpicks.

You said there may be an issue with resetting the backup state, so I'll wait for your go-ahead on that to approve.

app/src/main/AndroidManifest.xml
```diff
@@ -7,13 +7,15 @@
...
         android:minSdkVersion="26"
-        android:targetSdkVersion="26"
+        android:targetSdkVersion="28"
         tools:ignore="GradleOverrides,OldTargetApi" />
 
     <uses-permission
         android:name="android.permission.BACKUP"
         tools:ignore="ProtectedPermissions" />
 
+    <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
```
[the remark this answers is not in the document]
grote: This is needed for the JobScheduler btw. so it can re-schedule backups after re-boot.

app/src/main/java/com/stevesoltys/backup/service/backup/BackupObserver.java
```diff
@@ -65,14 +59,6 @@ public void backupPackageCompleted(BackupSession backupSession, String packageNa
 
     @Override
     public void backupSessionCompleted(BackupSession backupSession, BackupResult backupResult) {
-        ConfigurableBackupTransport backupTransport = ConfigurableBackupTransportService.getBackupTransport();
-
-        if (!backupTransport.isActive()) {
```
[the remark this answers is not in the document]
grote: I looked briefly for a way to know that a backup is already running, but didn't find anything. Since the backup itself are performed by the BackupManager, I would expect it to not start a second one while the first one is still running.

Later, I want to add a "Backup Now" button again, so then we should definitely look into this again.

app/src/main/java/com/stevesoltys/backup/transport/component/provider/ContentProviderRestoreComponent.java
grote: I hadn't seen `ContentProviderBackupConstants` and agree that the constants should go there!

PR #31, opened 2019-07-31
Android.mk: use prebuilt apk

chirayudesai:
After much deliberation, and multiple approaches towards a solution for building this with the AOSP build system, it seems that the best way is to just use a prebuilt, due to the usage of external libraries and Kotlin.

Added a script to download the apk from GitHub releases as well.


stevesoltys:
Thanks for the help!

PR #32, opened 2019-08-01
Android.mk: Actually make it work

chirayudesai:
I forgot some things.

And upon testing with the latest code (not the latest release), I found it doesn't compile, hence the pre-optimization disabling.


chirayudesai:
@stevesoltys My bad, you could squash the first two commits (required modules and privileged app) into yesterday's commit.

The last one could be separate because that's new and it's good to know why it's being done (Kotlin), and we may be able to figure out a solution to making it work.
Note: Dex pre-opt isn't enabled by default in most ROM builds, you need to `export WITH_DEXPREOPT=true` to turn it on, which we do at Calyxos as it makes sense for release builds.

PR #33, opened 2019-08-03
Release improvements

stevesoltys:
1. Remove XML configuration files from releases
2. Generate release for each commit and download during build.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -21,7 +21,7 @@ include $(CLEAR_VARS)
 backup_root := $(LOCAL_PATH)
 
 $(backup_root)/Backup.apk:
-	cd $(backup_root) && ./download.sh
+    curl -L https://github.com/stevesoltys/backup/releases/download/76bb8cd/app-release-unsigned.apk > $(backup_root)/Backup.apk
 
 LOCAL_MODULE := Backup
 LOCAL_SRC_FILES := Backup.apk
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -23,9 +23,14 @@ android {
     lintOptions {
         abortOnError false
     }
+
     compileOptions {
         targetCompatibility 1.8
         sourceCompatibility 1.8
+    }
+
+    dexOptions {
+        jumboMode = true
     }
 
     // optional signingConfigs
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,10 @@ cache:
     - $HOME/.gradle/wrapper/
     - $HOME/.android/build-cache
 
-before_deploy:
-  - git config --local user.name "Steve Soltys"
-  - git config --local user.email "github@stevesoltys.com"
-  - if ! [[ $TRAVIS_TAG ]]; then git tag $(git rev-parse --short HEAD); fi
+#before_deploy:
+#  - git config --local user.name "Steve Soltys"
+#  - git config --local user.email "github@stevesoltys.com"
+#  - if ! [[ $TRAVIS_TAG ]]; then git tag $(git rev-parse --short HEAD); fi
 deploy:
   provider: releases
   api_key:
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,10 @@ cache:
     - $HOME/.gradle/wrapper/
     - $HOME/.android/build-cache
 
+before_deploy:
+  - git config --local user.name "Steve Soltys"
+  - git config --local user.email "github@stevesoltys.com"
+  - if ! [[ $TRAVIS_TAG ]]; then git tag $(git rev-parse --short HEAD); fi
 deploy:
   provider: releases
   api_key:
@@ -33,5 +37,5 @@ deploy:
     - app/build/outputs/apk/release/app-release-unsigned.apk
   on:
     repo: stevesoltys/backup
-    tags: true
+    all_branches: true
   skip_cleanup: true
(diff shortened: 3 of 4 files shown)


chirayudesai:
Looks good to me.

Android.mk
```diff
@@ -21,7 +21,7 @@ include $(CLEAR_VARS)
 backup_root := $(LOCAL_PATH)
 
 $(backup_root)/Backup.apk:
-	cd $(backup_root) && ./download.sh
```
grote: Was this change intentional? Looks like it always downloads the same release?

stevesoltys: I was testing the build with a particular release to see if it would resolve the issue I'm having with this. It doesn't seem to be working with my build setup:

```
[  0% 47/12814] target Prebuilt: Backup (out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk)
FAILED: out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk 
/bin/bash -c "(rm -f out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk ) && (cp \"packages/apps/Backup/Backup.apk\" \"out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk\" ) && (if (zipinfo out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk 'lib/*.so' 2>/dev/null | grep -v ' stor ' >/dev/null) ; then rm -rf out/target/product/sailfish/obj/APPS/Backup_intermediates/uncompressedlibs && mkdir out/target/product/sailfish/obj/APPS/Backup_intermediates/uncompressedlibs; unzip -q out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk 'lib/*.so' -d out/target/product/sailfish/obj/APPS/Backup_intermediates/uncompressedlibs && zip -qd out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk 'lib/*.so' && ( cd out/target/product/sailfish/obj/APPS/Backup_intermediates/uncompressedlibs && find lib -type f | sort | zip -qD -X -0 ../package.apk -@ ) && rm -rf out/target/product/sailfish/obj/APPS/Backup_intermediate
 s/uncompressedlibs; fi ) && (if (zipinfo out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk '*.dex' 2>/dev/null | grep -v ' stor ' >/dev/null) ; then tmpdir=out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk.tmpdir; rm -rf \$tmpdir && mkdir \$tmpdir; unzip -q out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk '*.dex' -d \$tmpdir && zip -qd out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk '*.dex' && ( cd \$tmpdir && find . -type f | sort | zip -qD -X -0 ../package.apk -@ ) && rm -rf \$tmpdir; fi ) && (mv out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk.unsigned ) && (prebuilts/jdk/jdk9/linux-x86/bin/java -Djava.library.path=out/host/linux-x86/lib64 -jar out/host/linux-x86/framework/signapk.jar build/target/product/security/platform.x509.pem build/target/product/security/platform.pk8  out/target/product/sailfish/obj/AP
 PS/Backup_intermediates/package.apk.unsigned out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk.signed ) && (mv out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk.signed out/target/product/sailfish/obj/APPS/Backup_intermediates/package.apk )"
Picked up _JAVA_OPTIONS: -XX:-UsePerfData
java.util.zip.ZipException: zip END header not found
	at java.base/java.util.zip.ZipFile$Source.zerror(ZipFile.java:1252)
	at java.base/java.util.zip.ZipFile$Source.findEND(ZipFile.java:1153)
	at java.base/java.util.zip.ZipFile$Source.initCEN(ZipFile.java:1160)
	at java.base/java.util.zip.ZipFile$Source.<init>(ZipFile.java:997)
	at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:960)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:216)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:148)
	at java.base/java.util.jar.JarFile.<init>(JarFile.java:324)
	at java.base/java.util.jar.JarFile.<init>(JarFile.java:295)
	at java.base/java.util.jar.JarFile.<init>(JarFile.java:275)
	at com.android.signapk.SignApk.main(SignApk.java:1026)
```

chirayudesai: I'm unable to reproduce this, and from the looks of it the apk is downloaded, but it isn't able to use it correctly or something else.

Q / Android 10 is here though, so I want to see if we can build it properly inline there before coming back to this again.

stevesoltys: That sounds good to me.

I may try with a separate prebuilt repo to see if that resolves the issue I'm having for now.

grote: Could you maybe also check if the APK that gets copied there is valid? `zip END header not found` sounds like it might be broken?

Also, does this happen only on this branch or also on the `develop` branch?

stevesoltys: This also happens on the `develop` branch. I'm guessing the APK is invalid when it's copied there, but I'm not sure why.

I'm building on AWS so I'll need to stop the instance from shutting down after a failure so I can look at the APK and confirm that.

grote: If it also happens on the other branch, we could still clean up this one, merge it and solve the APK issue separately as it isn't related to this PR, right?

app/build.gradle
```diff
@@ -23,11 +23,16 @@ android {
...
+
     compileOptions {
         targetCompatibility 1.8
         sourceCompatibility 1.8
     }
 
+    dexOptions {
+        jumboMode = true
+    }
+
```
grote: Do we need this? What problem does it solve?

stevesoltys: It's been a bit since I did this, but I believe this was necessary since we pull in more dependencies now. I think I was experiencing some issues during the build?

Android.mk
stevesoltys: If we end up going with a separate prebuilt repo (i.e.: it's difficult to get this working with my build setup) then I don't want to move forward with this since it will create a GitHub release for every commit moving forward.

app/build.gradle
grote: Hmm.. at least here I don't run into any issues when not using jumboMode.

chirayudesai:
I tried to build it inline in Q - https://github.com/chirayudesai/backup/commit/b16dcf61f94c09f6f9d4f5833f8384be63a875a4 - but it doesn't work.

I don't see kotlinx in Q, and the versions of the androidx libraries differ too so at this point it seems like this will be another "keep adding prebuilt libraries" approach if we go for that.

In other news, the download.sh approach won't work anymore either since Q marks the source dir as read only and we can't write any files.

Prebuilt repo is the only approach I can think of which would work without any issues.

stevesoltys:
Gotcha @chirayudesai. I'll update this PR to move towards the prebuilt repo.

PR #46, opened 2019-10-27
Rebrand application

stevesoltys:
This rebrands the application. The new name will be "Seedvault".


app/src/main/AndroidManifest.xml
```diff
@@ -25,7 +25,7 @@
         tools:ignore="ProtectedPermissions" />
 
     <application
-        android:name=".Backup"
+        android:name=".Seedvault"
```
chirayudesai: We could just leave this as it is.

If not, the class needs to be renamed.

chirayudesai:
Works ok once the above comment is addressed.

PR #48, opened 2019-11-30
Update README.md to reflect latest features and permissions

stevesoltys:


grote:
Looks good!

dan-v:
@stevesoltys or @grote - I attempted an AOSP build with [seedvault-prebuilt](https://github.com/stevesoltys/seedvault-prebuilt) and it appears to be installed (I see it as a system app), but if I go to Settings->System->Backup it shows 'Off' (I'm assuming it should get registered there?). Is there a step that I'm missing? I also played around with setting `def_backup_enabled` to true and `def_backup_transport` to 'com.stevesoltys.seedvault/.transport.ConfigurableBackupTransportService' in [SettingsProvider](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/packages/SettingsProvider/res/values/defaults.xml#54) but no luck with that either.

grote:
@chirayudesai should know how. Otherwise, you should be able to set the transport with `adb bmgr`.

If that doesn't work, check logcat for errors.

dan-v:
Thanks @grote. The follow commands worked for me:
```
bmgr enable true
bmgr transport com.stevesoltys.seedvault.transport.ConfigurableBackupTransport
```

Thanks for your work on these changes! This is great.

stevesoltys:
Great to hear @dan-v.

@grote, maybe this warrants an issue, but I recall our transport being set automatically by the application in the past. Do you think we should re-instate that functionality so users don't need to use ADB? Or maybe this should be handled by setting this in AOSP before building?

grote:
The old version could activate itself, because it had an independent UI and thus code that could be run by the user. The new version is integrated in the settings, but only shows up there once selected. 

IMHO it is the OEM's responsibility to activate/select the transport. We should however document in the README (here or in prebuilt?) how to do it.

@chirayudesai  please correct me if this is wrong:

1.    overlay `frameworks/base/packages/SettingsProvider/res/values/defaults.xml` string `def_backup_transport` with `com.stevesoltys.seedvault.transport.ConfigurableBackupTransport` - for a new install.
1.    Bump database version in `com/android/providers/settings/DatabaseHelper.java` and set this string - for existing installs.

chirayudesai:
@grote You are correct, just the first part should work, I don't know why it didn't work for @dan-v 

Dan, did you check the output of `adb shell bmgr list transports` after you set that?

Also, that second step is what I wanted to do for existing users. That and actually changing the default transport provider via that db so that even upgrades get this default set.

I'll update with a link to the patch here / add it to the README once I have it working.

chirayudesai:
https://gitlab.com/calyxos/vendor_calyx/commit/c8bb7865bcde5d9c72fed4b84f3b004561b04fb6

This commit will configure the transport for new installs.

https://gitlab.com/calyxos/platform_frameworks_base/commit/9738a0c752dd75b1c6146882e2ca53a2232e7148

This commit will configure it for upgrades to existing installs.

PR #54, opened 2019-12-12
Support restore from Nextcloud account even if not installed or set up

grote:
This MR adds support for restoring a backup from a NextCloud account. If NextCloud is not installed, it allows the user to install it first. If there's no NextCloud account set up, it allows the user to set one up.

It also fixes a small bug that did not let the user escape when they forgot their recovery code.


app/src/main/java/com/stevesoltys/seedvault/transport/backup/plugins/DocumentsStorage.kt
```diff
@@ -126,3 +133,70 @@ fun DocumentFile.deleteContents() {
...
+ */
+@Throws(IOException::class)
+fun DocumentFile.listFilesBlocking(context: Context): ArrayList<DocumentFile> {
+    val resolver = context.contentResolver
+    val childrenUri = buildChildDocumentsUriUsingTree(uri, getDocumentId(uri))
+    val projection = arrayOf(COLUMN_DOCUMENT_ID, COLUMN_MIME_TYPE)
+    val result = ArrayList<DocumentFile>()
+
+    @SuppressLint("Recycle")  // gets closed in with(), only earlier exit when null
+    var c = resolver.query(childrenUri, projection, null, null, null)
```
stevesoltys: Maybe could use a more descriptive variable name here?

grote: Nowadays, the IDE shows you that it is a `Cursor` but sure, I changed the name.

PR #55, opened 2019-12-15
Update license and changelog

stevesoltys:


grote:
Looks good! BIP39 is not really the encryption, but the passphrase selection.

PR #64, opened 2020-01-09
Optional App APK Backups

### LINKED ISSUE #18: Add support for bundling application APKs
The application should allow users to select APKs to back up / restore.

grote:
This MR introduces our own backup metadata that is then used to allow the user to also back up (ordinary) APKs and have them restored for a better restore experience. It also improves the restore UI to show the restore state of each package.

Fixes #18, #65


stevesoltys:
@grote let me know when you feel this is complete.

stevesoltys:
Also, if possible, please point this to `develop` since any build on `master` will override the previous release version in the prebuilt repo (would like to only have releases in `master`).

grote:
>  let me know when you feel this is complete.

I thought it is complete, but now I am adding #65 to it.

> please point this to develop

Ah good to know that this is the workflow you envision.

grote:
If we don't find any bugs, this should be complete for now.

stevesoltys:
>Ah good to know that this is the workflow you envision.

Just a thought so people pointing to master on the prebuilt repo always get stable builds. I'm still open to suggestions if you have any that may improve the workflow.

PR #67, opened 2020-01-15
Improve UX for auto-restore when using removable storage

### LINKED ISSUE #51: Improve UX for auto-restore when using removable storage
Auto-restore restores app data when an app is re-installed and had a backup available. When backups are stored on removable storage (something not envisoned/supported by Google) the backup is not available to be restored. Though obvious to technical people, not all users might be aware of it.

So maybe we should explain this to users either once (when setting up removable storage) or every time an app is installed that we have a backup for on removable storage maybe with a heads up notification.

grote:
This MR also removes wildcard imports and improves the progress notification.

Closes #51


app/src/main/java/com/stevesoltys/seedvault/restore/RestoreErrorBroadcastReceiver.kt
```diff
@@ -0,0 +1,34 @@
...
+    // using KoinComponent would crash robolectric tests :(
+    private val notificationManager: BackupNotificationManager by lazy { get().koin.get<BackupNotificationManager>() }
+
+    override fun onReceive(context: Context, intent: Intent) {
+        if (intent.action != ACTION_RESTORE_ERROR_UNINSTALL) return
+
+        notificationManager.onRestoreErrorSeen()
+
+        val packageName = intent.getStringExtra(EXTRA_PACKAGE_NAME)!!
+        @Suppress("DEPRECATION")  // the alternative doesn't work for us
```
stevesoltys: Just curious, what is deprecated and what's the alternative?

grote: It is described [in the docs](https://developer.android.com/reference/android/content/Intent#ACTION_UNINSTALL_PACKAGE).

[The alternative](https://developer.android.com/reference/android/content/pm/PackageInstaller.html#uninstall(java.lang.String,%20android.content.IntentSender)) works only for:
*    the current "installer of record" for the package
*    the device owner
*    the affiliated profile owner 

and we are neither.

PR #76, opened 2020-03-08
Remove wrongly inserted "g" char from AndroidManifest.xml

ypid:
It was not caught by CI tests because it is still valid XML. Not even `gradle check` complained.

Introduced in: 78e217c7d8b4d0508b243a81b20417f7d9b9250e


stevesoltys:
Whoops... thank you!

Does this make the application unusable? If so, I will need to do a 1.0.1 release.

ypid:
You are very welcome!

> Does this make the application unusable?

I have not tested E2E but I assume not. Gradle will complain on broken XML. But your change was valid XML. Not sure how Android will behave.

PR #90, opened 2020-07-02
String improvements

mikeNG:
* Don't use camel case, following AOSP applications
* Set app name and Nextcloud as untranslatable


chirayudesai:
The untranslatable=false will make translations easier.

https://hosted.weblate.org/projects/calyxos/seedvault/

stevesoltys:
LGTM

PR #92, opened 2020-07-06
Strings reworked

comradekingu:


stevesoltys:
Please re-point this PR to ’develop’ (we work on that branch until we are ready for a release).

app/src/main/res/values/strings.xml
```diff
@@ -1,58 +1,58 @@
...
 
     <string name="restore_backup_button">Restore backup</string>
 
     <!-- Settings -->
     <string name="settings_backup">Backup my data</string>
-    <string name="settings_backup_location">Backup location</string>
-    <string name="settings_backup_location_picker">Choose backup location</string>
-    <string name="settings_backup_location_title">Backup Location</string>
-    <string name="settings_backup_location_invalid">The chosen location can not be used.</string>
+    <string name="settings_backup_location">Backup folder</string>
```
chirayudesai, line 10: We call this location because you don't actually get to select a particular folder to store the backup, just the location.

app/src/main/res/values/strings.xml
```diff
@@ -1,58 +1,58 @@
...
+    <string name="settings_backup_location_picker">Choose where to save the backup</string>
+    <string name="settings_backup_location_title">Backup folder</string>
+    <string name="settings_backup_location_invalid">The chosen folder can not be used.</string>
     <string name="settings_backup_location_none">None</string>
-    <string name="settings_backup_location_internal">Internal Storage</string>
+    <string name="settings_backup_location_internal">Internal storage</string>
     <string name="settings_backup_last_backup_never">Never</string>
-    <string name="settings_info">All backups are encrypted on your phone. To restore from backup you will need your 12-word recovery code.</string>
+    <string name="settings_info">All backups are encrypted by your device. To restore one you need your 12-word recovery code.</string>
```
chirayudesai, line 17: I would keep on device, given that they're encrypted on the device itself.

Also perhaps "to restore it" instead of "one"

app/src/main/res/values/strings.xml
```diff
@@ -1,58 +1,58 @@
...
     <string name="settings_auto_restore_title">Automatic restore</string>
-    <string name="settings_auto_restore_summary">When reinstalling an app, restore backed up settings and data.</string>
-    <string name="settings_auto_restore_summary_usb">Note: Your %1$s needs to be plugged in for this to work.</string>
+    <string name="settings_auto_restore_summary">When recovering an app, also include settings and data.</string>
```
chirayudesai, line 19: reinstalling is the right term here, since all you need to do is re-install the app, you don't specfiically restore it, that's done automatically.

app/src/main/res/values/strings.xml
```diff
@@ -67,29 +67,29 @@
...
     <string name="notification_title">Backup running</string>
-    <string name="notification_backup_result_complete">Backup complete</string>
-    <string name="notification_backup_result_rejected">Not backed up</string>
-    <string name="notification_backup_result_error">Backup failed</string>
+    <string name="notification_backup_result_complete">Done</string>
+    <string name="notification_backup_result_rejected">Could not perform backup</string>
+    <string name="notification_backup_result_error">Try again</string>
 
-    <string name="notification_success_title">Backup finished</string>
+    <string name="notification_success_title">Done</string>
```
chirayudesai, line 80: Mentioning the backup makes this more clearer

app/src/main/res/values/strings.xml
```diff
@@ -67,29 +67,29 @@
...
     <string name="notification_success_num_not_backed_up">%1$d apps could not get backed up</string>
-    <string name="notification_failed_title">Backup failed</string>
+    <string name="notification_failed_title">That didn\'t go well</string>
```
chirayudesai, line 82: Same here.

app/src/main/res/values/strings.xml
```diff
@@ -67,29 +67,29 @@
...
 
-    <string name="notification_error_channel_title">Error Notification</string>
-    <string name="notification_error_title">Backup Error</string>
-    <string name="notification_error_text">A device backup failed to run.</string>
+    <string name="notification_error_channel_title">Error notification</string>
+    <string name="notification_error_title">Backup error</string>
+    <string name="notification_error_text">One device was not backed up.</string>
```
chirayudesai, line 86: One doesn't sound right here, since it's not like you can backup multiple devices.

"Backup failed to run." perhaps

app/src/main/res/values/strings.xml
```diff
@@ -98,31 +98,31 @@
...
     <string name="restore_app_quota_exceeded">Backup quota exceeded</string>
-    <string name="restore_finished_success">Restore complete</string>
-    <string name="restore_finished_error">An error occurred while restoring the backup.</string>
+    <string name="restore_finished_success">Restored</string>
+    <string name="restore_finished_error">Could not restore the backup.</string>
     <string name="restore_finished_button">Finish</string>
     <string name="storage_internal_warning_title">Warning</string>
-    <string name="storage_internal_warning_message">You have chosen internal storage for your backup. This will not be available when your phone is lost or broken.</string>
-    <string name="storage_internal_warning_choose_other">Choose Other</string>
+    <string name="storage_internal_warning_message">You have chosen internal storage for your backup, and it will not help you if your phone is lost or broken.</string>
```
chirayudesai, line 115: not be available is more clearer than "not help you" IMO

chirayudesai (changes requested):
Left some inline comments, the other improvements look good.

Could you also rebase this on develop so that we just have this commit and the changes you made instead of the merges too.

Thanks!

app/src/main/res/values/strings.xml
comradekingu: How about "Store backups in"?
Not knowing the distinction between "Backup" and "Back up" could otherwise cause confusion

app/src/main/res/values/strings.xml
comradekingu, line 17:
```suggestion
    <string name="settings_info">All backups are encrypted by your device, to be restored by your 12-word recovery code.</string>
```
I tried to avoid "on" possibly meaning "onto". "By" still allows the backup to be stored elsewhere.
By "One", I tried to reference "device", or "backup", can't remember. That failed.

app/src/main/res/values/strings.xml
comradekingu, line 86:
```suggestion
    <string name="notification_error_text">Could not perform backup.</string>
```
"Backup failed to run." is quirky English for no good reason.

app/src/main/res/values/strings.xml
comradekingu: I argue the user needs to be alerted to how helpless that situation is. Not only isn't it available on your device, it isn't available anywhere short of possible recovery.

comradekingu, line 115:
```suggestion
    <string name="notification_error_text">Could not perform backup.</string>
```
"Backup failed to run." is quirky English for no good reason.
```suggestion
    <string name="storage_internal_warning_message">You have chosen internal storage for your backup, and it will not help you if your device is lost or broken.</string>
```

app/src/main/res/values/strings.xml
comradekingu: True, but is that extra info ever needed? It is slower.
My understanding is that nothing is wrong, the user knows what it attempted to do, which can now just be acknowledged.

app/src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="notification_failed_title">Could not perform backup</string>
```

app/src/main/res/values/strings.xml
comradekingu: My confusion: "Reinstalling" could mean the app would have to be re-downloaded, whereas "recovering", means the app is already there, and that in addition to this, the settings and data _could_ be included in the recovery of it.

In rare cases, I think confusion could arise over whether "settings and data" is fresh data, seeing as "recovery" is more closely tied to something existing, than "reinstalling", which could just overwrite it.

app/src/main/res/values/strings.xml
chirayudesai: I'm not sure I understand what you mean by "Backup" and "Back up" here.

app/src/main/res/values/strings.xml
chirayudesai: It does have to be redownloaded, that is what this setting is for.

The way it works
1) You have a backup containing app AwesomeMusicPlayer
2) You have since uninstalled it
3) At a later date, you re-install AwesomeMusicPlayer (redownload needed)
4) If the backup is available, Seedvault will try to restore it's data automatically. hence, Automatic restore.

app/src/main/res/values/strings.xml
chirayudesai: It doesn't hurt just having it there though IMO

app/src/main/res/values/strings.xml
chirayudesai: Sure, then we can make it more clearer, and not more opaque.

app/src/main/res/values/strings.xml
comradekingu: Backup, noun. Back up, verb.

app/src/main/res/values/strings.xml
comradekingu: That is why I put in "by"… Otherwise it could strictly be that they were in the state of encrypted, on the device. Since the backups are not marked in sentence 1, and unreferenced "one" in sentence 2 fits.

app/src/main/res/values/strings.xml
comradekingu: I got confused with "re-install" meaning it not only has, but _is_ installed beforehand. I could be wrong.

app/src/main/res/values/strings.xml
comradekingu: "Backup failed to run" is very non-descriptive. The result as plainly as possible is what matters.

app/src/main/res/values/strings.xml
comradekingu, line 80:
```suggestion
    <string name="notification_success_title">Backup done</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -1,58 +1,58 @@
...
     <string name="settings_backup_location_none">None</string>
-    <string name="settings_backup_location_internal">Internal Storage</string>
+    <string name="settings_backup_location_internal">Internal storage</string>
     <string name="settings_backup_last_backup_never">Never</string>
-    <string name="settings_info">All backups are encrypted on your phone. To restore from backup you will need your 12-word recovery code.</string>
+    <string name="settings_info">All backups are encrypted by your device, to be restored by your 12-word recovery code.</string>
     <string name="settings_auto_restore_title">Automatic restore</string>
-    <string name="settings_auto_restore_summary">When reinstalling an app, restore backed up settings and data.</string>
-    <string name="settings_auto_restore_summary_usb">Note: Your %1$s needs to be plugged in for this to work.</string>
+    <string name="settings_auto_restore_summary">When recovering an app, also include settings and data.</string>
```
comradekingu, line 19:
```suggestion
    <string name="settings_auto_restore_summary">When installing a previously installed app, recover its settings and data.</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -1,58 +1,58 @@
...
     <string name="settings_backup_location_none">None</string>
-    <string name="settings_backup_location_internal">Internal Storage</string>
+    <string name="settings_backup_location_internal">Internal storage</string>
     <string name="settings_backup_last_backup_never">Never</string>
-    <string name="settings_info">All backups are encrypted on your phone. To restore from backup you will need your 12-word recovery code.</string>
+    <string name="settings_info">All backups are encrypted by your device, to be restored by your 12-word recovery code.</string>
     <string name="settings_auto_restore_title">Automatic restore</string>
-    <string name="settings_auto_restore_summary">When reinstalling an app, restore backed up settings and data.</string>
-    <string name="settings_auto_restore_summary_usb">Note: Your %1$s needs to be plugged in for this to work.</string>
+    <string name="settings_auto_restore_summary">When installing a previously installed app, recover its settings and data.</string>
```
comradekingu, line 19:
```suggestion
    <string name="settings_auto_restore_summary">When installing a previously uninstalled app, recover its settings and data.</string>
```

chirayudesai:
Re-pointed to 'master' since we got rid of 'develop'

chirayudesai:
@mikeNG do you mind taking a look?

@comradekingu https://calyxinstitute.org/projects/seedvault-encrypted-backup-for-android has a video of the initial version if that helps.

This would need to be rebased to fix conflicts, and also squashed, but that's before merging, can wait.

chirayudesai:
Closing since it's gone a while without any updates.

PR #105, opened 2020-07-27
add requirements topic

steadfasterX:
let users know that A10 is needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ A backup application for the [Android Open Source Project](https://source.androi
 - Automatic daily backups that run in the background.
 
 ## Requirements
-- Android 10 or higher
+- Android 10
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,6 +9,9 @@ A backup application for the [Android Open Source Project](https://source.androi
 - User-friendly encryption using a mnemonic phrase (BIP39).
 - Automatic daily backups that run in the background.
 
+## Requirements
+- Android 10 or higher
```
stevesoltys, line 13: Maybe just "Android 10" since we may not support 11 when it is released?

steadfasterX: sounds reasonable yes

grote:
Thanks @steadfasterX!

Issue #95, opened 2020-07-09
Improve performance and reliability when using Nextcloud as storage location


grote:
Optimize the interactions with Android's storage access framework (SAF) so that the number of network requests performed by Nextcloud is minimized to improve backup speed and reliability.

* [x] write (on-device) unit tests for the current storage abstraction [[PR](https://github.com/stevesoltys/seedvault/pull/107)]
* [x] replace usage of DocumentFile abstraction with fine-grained ContentProvider queries (was not beneficial, instead we eliminated parallelism as much as possible on our side) [[PR](https://github.com/stevesoltys/seedvault/pull/109)]
* [x] test and benchmark the effects against a Nextcloud test-instance
* [x] improve notification progress bar to account for extremely slow start when using Nextcloud [[PR](https://github.com/stevesoltys/seedvault/pull/111)]

beerisgood:
If you need testing, I can help.

What I already can say is that using pin in Nextcloud app prevent sync but that may be Nextcloud (app) related.

PR #98, opened 2020-07-09
Make Seedvault compile in AOSP properly

mikeNG:
* Fixes #97

PR #107, opened 2020-08-06
Add instrumentation tests for storage plugin (SAF)

grote:
First part of  #95

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
@@ -49,6 +49,7 @@ class SettingsManager(context: Context) {
         return Storage(uri, name, isUsb)
     }
 
+    // TODO find a better solution for this hack abusing the settings manager
     fun getAndResetIsStorageChanging(): Boolean {
         return isStorageChanging.getAndSet(false)
     }
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,8 @@
 
 buildscript {
 
+    // 1.3.21 Android 10
+    // 1.3.72 AOSP master (2020-08)
     ext.kotlin_version = '1.3.61'
 
     repositories {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -129,14 +129,17 @@ dependencies {
     lintChecks 'com.github.thirdegg:lint-rules:0.0.4-alpha'
 
     def junit_version = "5.5.2"
+    def mockk_version = "1.10.0"
     testImplementation aospDeps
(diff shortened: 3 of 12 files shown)


app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
```diff
@@ -49,6 +49,7 @@ class SettingsManager(context: Context) {
         return Storage(uri, name, isUsb)
     }
 
+    // TODO find a better solution for this hack abusing the settings manager
```
stevesoltys, line 52: Maybe this could be a research task.

build.gradle
```diff
@@ -2,7 +2,7 @@
 
 buildscript {
 
-    ext.kotlin_version = '1.3.61'
+    ext.kotlin_version = '1.3.72'
```
chirayudesai, line 5: Need to test this with #98

grote: Might be good to know what Kotlin version AOSP10 uses so we can stay in sync with that.

chirayudesai:
Is "Upgrade gradle and kotlin" needed for this PR? If not, could merge without that, or just the gradle update.

build.gradle
chirayudesai: 1.3.21 in the latest Android 10 release (august)
1.3.72 in master, let's see what R / 11 has.

Source: https://android.googlesource.com/platform/external/kotlinc/+/refs/tags/android-10.0.0_r41/build.txt

grote: Hmm... so maybe it is better to downgrade to that? I *think* that all 1.3.x should be compatible, but who knows...

grote:
> Is "Upgrade gradle and kotlin" needed for this PR? If not, could merge without that, or just the gradle update.

The gradle update shouldn't be an issue as it isn't used for AOSP builds. But we should decide what we do with the kotlin compiler version.

build.gradle
chirayudesai: Not sure. So far it hasn't been an issue, but yeah might be better to stay in sync.

But then again 11 isn't too far out, so let's decide then?

chirayudesai:
> > Is "Upgrade gradle and kotlin" needed for this PR? If not, could merge without that, or just the gradle update.
> 
> The gradle update shouldn't be an issue as it isn't used for AOSP builds. But we should decide what we do with the kotlin compiler version.

Yep, so could you drop the kotlin bit from this commit/PR for now, we can deal with that later.

build.gradle
grote: Ok so I reverted this change and left the Kotlin version as it was.

chirayudesai:
Thanks!

PR #108, opened 2020-08-07
Merge master into develop

stevesoltys:


grote:
Do we really need this in `develop`?

stevesoltys:
Personally, I believe develop should always be up to date with master.

grote:
Yeah I am just wondering about the process here as I don't really like these reverse PRs, especially for small changes. One option is to let PRs always target develop (or use a different stable branch and always target master/main). Another is to re-rebase develop on master, but that might cause issues with open PRs.

stevesoltys:
Hmm, why don't you like them?

I would think the best solution would be to cascade merges made to the latest stable version (`master`) into the latest development version (`develop`). Bitbucket [has this functionality](https://confluence.atlassian.com/bitbucketserver/automatic-branch-merging-776639993.html) and I'm sure it's possible with GitHub actions as well. Do you think you'd be okay with that?

grote:
> Hmm, why don't you like them?

It creates unnecessary merge commits which clutter the history (especially when you eventually merge them back into `master`).

After thinking about it for a bit, my preferred flow would be to have a dedicated stable branch (could be named after the release we currently target, e.g. `android10`) that people build their releases from and use `master` (or `main`) for ongoing development.
This has the advantage that there's no confusion about which branch to target with PRs. You then would never directly push to the stable branch (except it becomes old-stable) and only ever merge from main/master to stable.

chirayudesai:
I like this idea, `android10` would be good - and we can simply leave it as-is if/when we have to add any breaking changes for future versions. If not, `stable` works too.

stevesoltys:
@grote @chirayudesai sure, that works as well.

chirayudesai:
Done, see https://github.com/stevesoltys/seedvault/pull/110

chirayudesai:
We gotta do it this way for all the status checks to be happy.

PR #114, opened 2020-09-04
Show notification for backup running in the background

grote:
The system triggers backup jobs periodically or when a package is announcing that its data has changed. So far we were not showing notifications for those.

This MR shows a notification with an indeterminate progress bar as we don't have any information about how many packages will get backed up.

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
@@ -198,6 +198,8 @@ class MetadataManager(
 
     @Synchronized
     fun getPackagesNumBackedUp(): Int {
+        // FIXME we are under-reporting packages here,
+        //  because we have no way to also include upgraded system apps
         return metadata.packageMetadataMap.filter { (_, packageMetadata) ->
             !packageMetadata.system && ( // ignore system apps
                     packageMetadata.state == APK_AND_DATA || // either full success
--- a/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
@@ -3,6 +3,8 @@ package com.stevesoltys.seedvault.settings
 import android.app.Application
 import android.content.pm.PackageManager.NameNotFoundException
 import android.util.Log
+import android.widget.Toast
+import android.widget.Toast.LENGTH_LONG
 import androidx.annotation.UiThread
 import androidx.core.content.ContextCompat.getDrawable
 import androidx.lifecycle.LiveData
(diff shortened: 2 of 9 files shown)


app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
```diff
@@ -198,6 +198,8 @@ class MetadataManager(
 
     @Synchronized
     fun getPackagesNumBackedUp(): Int {
+        // FIXME we are under-reporting packages here,
+        //  because we have no way to also include upgraded system apps
```
chirayudesai, line 202: spacing

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
```diff
@@ -65,7 +69,11 @@ internal class SettingsViewModel(
     }
 
     internal fun backupNow() {
-        Thread { requestBackup(app) }.start()
+        if (notificationManager.hasActiveBackupNotifications()) {
+            Toast.makeText(app, R.string.notification_backup_already_running, LENGTH_LONG).show()
```
chirayudesai, line 73: I'm not the biggest fan of toast messages tbh.

A snackbar might be better.

app/src/main/java/com/stevesoltys/seedvault/ui/notification/BackupNotificationManager.kt
```diff
@@ -16,6 +16,7 @@ import androidx.core.app.NotificationCompat.Builder
 import androidx.core.app.NotificationCompat.PRIORITY_DEFAULT
 import androidx.core.app.NotificationCompat.PRIORITY_HIGH
 import androidx.core.app.NotificationCompat.PRIORITY_LOW
+import com.stevesoltys.seedvault.BuildConfig.APPLICATION_ID
```
chirayudesai, line 19: This won't be available when building in AOSP, without gradle.

Looks like it's already in the other notification commit too - https://github.com/stevesoltys/seedvault/pull/111 - that'll break build.

chirayudesai (changes requested):

app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
grote: The spacing looks indeed odd, but it is intentional. This way the IDE highlights the multi line comment as FIXME

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
grote: We are in a view model here and can't show a snackbar. This is just to give some sort of feedback why we won't run a backup now and under normal circumstances, this won't be triggered (only when a user wants to manually trigger a backup, but didn't notice that one is already running). Not sure the extra effort for a snackbar would be justified.

chirayudesai:
Some alternatives in case `BuildConfig` is needed for other things in the future:
* https://cs.android.com/search?q=BuildConfig.java&sq=
* https://github.com/LineageOS/android_packages_apps_Eleven/commit/aff7e1d73f82ab2d8dfb1a08ef469db32422fb8c

Basically just commit a BuildConfig.java for aosp build in a separate dir

chirayudesai (approved):

stevesoltys (approved):

PR #120, opened 2020-09-09
String improvements

chirayudesai:
* Don't use camel case, following AOSP applications
* Set app name and Nextcloud as untranslatable


chirayudesai:
This is only if we don't merge develop into android10 right away.

chirayudesai:
Should start to merge translations once this is in.

chirayudesai:
We're just gonna merge develop wholesale once #124 is in

chirayudesai:
Yeah let's just stick with merging master to keep things simpler.

If there are any issues we fix it, like we did #128

PR #121, opened 2020-09-09
Android.bp support for the stable branch

chirayudesai:


chirayudesai:
This is only if we don't merge develop into android10 right away.

chirayudesai:
If this is merged https://github.com/stevesoltys/seedvault/pull/119 should be picked too.

I can just put that in this branch.

chirayudesai:
We're just gonna merge develop wholesale once #124 is in

chirayudesai:
Let's just stick with merging master to keep things simpler.

If there are any issues we fix it, like we did #128

PR #110, opened 2020-08-26
Merge develop into master

chirayudesai:
I have already created an android10 branch which has the contents of current master.

This branch will now be the primary development branch.
Old master aka current android10 will become the stable branch.
android11 will be created when that's out and the targeted release, and that will make it easier to do breaking changes if ever needed.

PR #124, opened 2020-09-10
Master merge develop

chirayudesai:
Take 4 for #110 

Hopefully this works


grote (approved):

grote:
Looks like this includes all commits from `develop`.

chirayudesai:
Explanation:

`master` used to be the stable branch, and `develop` the development branch.
We had one PR merged into `master` - a super small README change, that wasn't in develop.
That sprouted a merge master into develop PR, #108 - but that just seems wrong, it should usually be the other way.
So we decided on that with #110 - and it took some time to get going.
But GitHub won't let us merge or push that, since it doesn't like merging diverging branches with all these fancy new checks.
It wouldn't even let me push - I know what I'm doing, just. let. me. git. push.

Anyway, I tried a bunch of stuff - #122 #123 but eventually settled on this, which is what I intended to git push but as a PR instead.
After this develop can be deleted and we can all live happily ever after.

stevesoltys (approved):

PR #128, opened 2020-09-14
Optimize Key-Value Backup

grote:
This PR caches the folder contents when making key-value backups or restores as it turned out getting files via SAF is super slow and can make operations with many records (e.g. `@pm@` and call log backup) very long.


stevesoltys (approved):

chirayudesai:
Why is it still called `app/src/main/`**`java`** when all we have is kotlin

Not that I'm suggesting a rename, just noticed it and found it amusing.

PR #131, opened 2020-09-18
Fix opt-out apps showing up as not yet backed up

grote:
This bug also caused APKs of opt-out apps not getting backed up.

This PR also shows a different text for stopped apps in app backups status page to make it clearer why they didn't get backed up. Since the fix involved introducing a dedicated state for this case, this was an easy win.
Also, when long tapping apps, the user is now brought to app system settings.

While testing, I noticed that manually triggered backups were not incremental, because a wrong integer flag was used which accidentally forced manual backups to be non-incremental.


app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -154,6 +154,11 @@ internal fun PackageInfo.isNotUpdatedSystemApp(): Boolean {
 
 internal fun PackageInfo.doesNotGetBackedUp(): Boolean {
     if (packageName == MAGIC_PACKAGE_MANAGER || applicationInfo == null) return true
-    return applicationInfo.flags and FLAG_ALLOW_BACKUP == 0 && // does not allow backup
+    return applicationInfo.flags and FLAG_ALLOW_BACKUP == 0 || // does not allow backup
```
grote, line 157: This two character change is the core of the fix. The rest is just resolving collateral damage from not properly differentiating between `FLAG_ALLOW_BACKUP` and `FLAG_STOPPED`.

chirayudesai:
> Some formatting fixes (to Kotlin official style guide) were made along the way.

We have discussed this before, but still I'm not sure what's worse - sneaking in formatting changes like this, or just doing it in one commit and getting it over with (including committing any style guide / lint settings if any)

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
chirayudesai: Worth splitting out in a separate commit perhaps?

chirayudesai (changes requested):
Let's leave out the formatting changes here and just fix it repo-wide once and for all.

grote:
> We have discussed this before, but still I'm not sure what's worse - sneaking in formatting changes like this, or just doing it in one commit and getting it over with (including committing any style guide / lint settings if any)

I think we should do it in one commit/PR for all files to get rid of this topic. However, I already have several other PRs in preparation, so that I would not like to do it right away, but I should be able to do it on top of the Android 11 PR without too much extra work.

> Let's leave out the formatting changes here and just fix it repo-wide once and for all.

Do you think it is worth doing the extra work of now undoing the formatting changes here manually instead of spending this time on something more useful?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
grote: I split out this commit and clarified the other changes in the commit message of the second commit.

chirayudesai:
> > We have discussed this before, but still I'm not sure what's worse - sneaking in formatting changes like this, or just doing it in one commit and getting it over with (including committing any style guide / lint settings if any)
> 
> I think we should do it in one commit/PR for all files to get rid of this topic. However, I already have several other PRs in preparation, so that I would not like to do it right away, but I should be able to do it on top of the Android 11 PR without too much extra work.

Sure, do it whenever it suits you, but sometime in 11 and we leave 10 as-is.

> 
> > Let's leave out the formatting changes here and just fix it repo-wide once and for all.
> 
> Do you think it is worth doing the extra work of now undoing the formatting changes here manually instead of spending this time on something more useful?

Normally, no, but since this is a bugfix it'd be worth it I think. I imagine your other commits done for 11 previously wouldn't have had the formatting changes so that should make things easier for those too from what you mentioned above, right?

stevesoltys (approved):
I'll leave my opinion here on the matter of formatting. Since most (all?) of the formatting changes are in files where the actual logic changes are made, I personally feel like it's okay to include them with this PR as is, since they are at least in some way related.

I do agree that a subsequent PR should apply formatting across the repository, if possible, to avoid this in the future. I will approve, but if it's still a pressing issue, we could continue the discussion.

stevesoltys (approved):

app/src/main/java/com/stevesoltys/seedvault/transport/ConfigurableBackupTransportService.kt
```diff
@@ -63,7 +63,7 @@ fun requestBackup(context: Context) {
     val observer = NotificationBackupObserver(context, packages.size, appTotals)
     val result = try {
         val backupManager: IBackupManager = get().koin.get()
-        backupManager.requestBackup(packages, observer, BackupMonitor(), FLAG_USER_INITIATED)
+        backupManager.requestBackup(packages, observer, BackupMonitor(), 0)
```
chirayudesai, line 66: @grote was this the issue with the flag and incremental backup you were talking about?

app/src/main/res/values/strings.xml
```diff
@@ -106,7 +106,8 @@
     <string name="restore_restoring">Restoring backup</string>
     <string name="restore_magic_package">System package manager</string>
     <!-- This text gets shown for apps that the OS did not try to backup for whatever reason e.g. no backup was run yet -->
-    <string name="restore_app_not_eligible">Not yet backed up</string>
+    <string name="restore_app_not_yet_backed_up">Not yet backed up</string>
```
chirayudesai, line 109: We should revisit this string later, see if we can come up with something better. Can be a part of a string improvement PR though.

chirayudesai (approved):
LGTM.

I only wanted to have it without the formatting changes just this once, since this is meant to go into android10 stable, and we're at android11 already, so the less changes we make there, the better.

Testing really does take up a lot of time with backup / restore, so anything that can help with that is appreciated.

Thanks for undoing the formatting changes Torsten.

app/src/main/java/com/stevesoltys/seedvault/transport/ConfigurableBackupTransportService.kt
grote: Yes, exactly. See also the commit message of this commit.

Issue #63, opened 2020-01-08
Better SetupWizard integration


chirayudesai:
This ticket intentionally has a vague sounding title.

We could allow the user to configure backups via the SetupWizard, that would be it's own thing.

And at the same time, we could try and make the design and feel of the current restore screen match the SetupWizard as much as possible. This would be useful for a future backup screen too.

It's nothing major, but I mean tiny things like making the "don't restore" and "next" button look more like they already do for other SetupWizard buttons texts.

Minor nits like these are what make it look less like a SetupWIzard page and more like another app.

grote:
>  Configure Backup from SetupWizard

This should probably be a ticket in the SetupWizard repo as well.

>  Make the restore screen's buttons more like SetupWizard

Or the other way around. Seedvault uses the normal colored material design buttons as the stock SetupWizard uses. See https://github.com/stevesoltys/seedvault/issues/34

>  Make the restore screen not fully immersive. Need to see the navbar and status bar.

Now resolved in https://github.com/grote/backup/commit/c52d3724426ae8b832dd5c4f5a937e5c44e9a58a can be checked.

grote:
> Configure Backup from SetupWizard

Google is showing a notification after setup (and periodically) that the phone doesn't get backed up, prompting the user to set up backup. This might be an alternative that doesn't keep the user writing down recovery codes during setup wizard.

chirayudesai:
> > Configure Backup from SetupWizard
> 
> Google is showing a notification after setup (and periodically) that the phone doesn't get backed up, prompting the user to set up backup. This might be an alternative that doesn't keep the user writing down recovery codes during setup wizard.

That could work. Could show notification on first boot, and clear it when backup is setup.

chirayudesai:
> - [x] Make the restore screen not fully immersive. Need to see the navbar **and** status bar.

I can't remember why it says navbar **and** status bar here, I guess I may have meant the in-app / suw nav bar rather than the system navbar.

Anyway, current / android10 behaviour is to hide the nav bar (but show the status bar), and that is correct, since then it looks like just another setupwizard page.

grote:
@chirayudesai can we check the second box now?

PR #133, opened 2020-09-25
Move to Android 11

grote:
After this PR has been merged into `master`, it will not build anymore within AOSP 10. People wanting to maintain Seedvault for Android 10 are welcome to submit PRs for the [`android10`](https://github.com/stevesoltys/seedvault/commits/android10) branch.

I've done all my testing and dev work on top of this PR for the past weeks. It is included in GrapheneOS 11 preview release. Therefore, I consider it stable and tested enough to be merged into `master`. After additional testing and confidence, we can branch off a stable `android11` branch from it.


chirayudesai:
I will push the current android10 to a 'master-android10' in my fork so that we can PR it for merge to android10 later.

I want to get translations merged too for android10, but that's not a blocker as they can be picked up separately without issue.

chirayudesai:
> I will push the current android10 to a 'master-android10' in my fork so that we can PR it for merge to android10 later.

Done, see #135 
> 
> I want to get translations merged too for android10, but that's not a blocker as they can be picked up separately without issue.

Done, #134 


This means we can merge this freely into master, and still have something to merge into android10 once tested.

app/src/test/java/com/stevesoltys/seedvault/metadata/MetadataManagerTest.kt
```diff
@@ -34,6 +35,7 @@ import java.io.IOException
 import kotlin.random.Random
 
 @RunWith(AndroidJUnit4::class)
+@Config(sdk = [29]) // robolectric does not support 30, yet
```
chirayudesai, line 39: Do we know when are they going to?

I think maybe there was a link to some upstream issue - would be nice to have that somewhere, can even be a separate issue here "Update roboelectric once upstream is fixed" or something.

grote: No idea. They've just got around to support 29 properly.

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreActivity.kt
```diff
@@ -19,8 +19,6 @@ class RestoreActivity : RequireProvisioningActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        if (isSetupWizard) hideSystemUI()
```
chirayudesai, line 22: Not needed anymore?

Android.bp
```diff
@@ -44,7 +44,10 @@ android_app {
     required: [
         "privapp_whitelist_com.stevesoltys.backup",
         "com.stevesoltys.backup_whitelist"
-    ]
+    ],
+    optimize: {
+        enabled: false,
```
chirayudesai, line 49: `java.lang.NoClassDefFoundError` was the issue right?

Could setup a custom `proguard.flags` in the future if needed.

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreActivity.kt
grote: I think you said that the system UI should still be shown in SUW, right? The code for doing that changed on Android 11.

chirayudesai (commented):
Did you try this with SetupWizard yet?

Asking due to the removal of the `SYSTEM_UI_FLAG_HIDE_NAVIGATION` bit.

LGTM.

Android.bp
grote: That actually had another reason, so we *could* re-enable proguard, but it is risky.

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreActivity.kt
chirayudesai: But I don't see any new code that still does this. Should make a note somewhere if this needs updating.

Android.bp
chirayudesai: Ah. I don't mind just leaving it disabled, definitely the better option for now.

chirayudesai (approved):

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreActivity.kt
grote: Yes due to #63 I didn't add the new code.

chirayudesai: Ah, I see the confusion, that shouldn't say *and* - we want it to look like what it currently does / has for the year so far.

chirayudesai (changes requested):
We just need to bring back https://github.com/stevesoltys/seedvault/commit/65905a7160cbae595aba3f05810532931042bbfb to undo the changes in the "Android 11 warnings" commit.

Can leave a note in the general Android 11 to update this to a newer API if available.

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreActivity.kt
grote: alright, I added it back.

chirayudesai:
Can you amend "Bring back hiding of navigation bar in SetupWizard" into "Fix compile warnings on AOSP 11"

Just makes it confusing otherwise when looking at history.

Untested personally, but code looks good and good to go otherwise.

chirayudesai (approved):
857f503a8fcd511136227ec0b38426a0c491d7f8 - the latest, is indentical to the commit before amending the setupwizard change - i.e. a7dc7f52d340243e254f129df958624329985af1

Thanks!

Edit: Identical as in `git diff` output is empty, the only change is that the new extra commit is gone and amended into "Fix compile warnings in AOSP 11"

PR #139, opened 2020-10-07
Auto-format code style to match official style and do CI check

grote:


.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,149 @@
...
+      <option name="LINE_COMMENT_AT_FIRST_COLUMN" value="false" />
+      <option name="LINE_COMMENT_ADD_SPACE" value="true" />
+      <option name="KEEP_BLANK_LINES_IN_DECLARATIONS" value="1" />
+      <option name="KEEP_BLANK_LINES_IN_CODE" value="1" />
+      <option name="KEEP_BLANK_LINES_BEFORE_RBRACE" value="1" />
+      <option name="ALIGN_MULTILINE_PARAMETERS" value="false" />
+      <option name="PARAMETER_ANNOTATION_WRAP" value="1" />
+      <option name="VARIABLE_ANNOTATION_WRAP" value="1" />
+      <option name="ENUM_CONSTANTS_WRAP" value="1" />
+      <indentOptions>
```
chirayudesai, line 144: Are the settings based on something specific somewhere? If so, we can add a link somewhere for easier reference in the future.

README.md
```diff
@@ -32,7 +32,9 @@ It uses the same internal APIs as `adb backup` which is deprecated and thus need
 * `android.permission.INSTALL_PACKAGES` to re-install apps when restoring from backup.
 
 ## Contributing
-Bug reports and pull requests are welcome on GitHub at https://github.com/stevesoltys/seedvault. 
+Bug reports and pull requests are welcome on GitHub at https://github.com/stevesoltys/seedvault.
+
+This project aims to adhere to the [official Kotlin coding style](https://developer.android.com/kotlin/style-guide).
```
chirayudesai, line 37: i.e. the previous question was more is there a reference codestyle file for this?

app/build.gradle
```diff
@@ -74,6 +77,17 @@ android {
 
 apply from: '../gradle/dependencies.gradle'
 
+ktlint {
+    version = "0.36.0" // https://github.com/pinterest/ktlint/issues/764
+    android = true
+    enableExperimentalRules = false
+    verbose = true
+    disabledRules = [
+            "import-ordering",
```
chirayudesai, line 86: Any reason for disabling this? Would it make too many changes?

.idea/codeStyles/Project.xml
grote: This comes from https://github.com/pinterest/ktlint#option-3 but not sure you can add a link to that file here. Anyway, now the CI will check if the style is ok and fail if not.

README.md
grote: Well, this here is the docs for Android which ktlint checks. As always there's some slight disagreements/bugs between the IDE and ktlint, so some smaller fixes are needed here and there. I am not aware of a reference xml Intellj file for the coding style. The most important bit there is

```xml
<option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
```

app/build.gradle
grote: The reason is that the IDE does not yet support the official ordering. So when you auto-format a file in the IDE, the CI will fail.

chirayudesai (commented):
Most of the changes are whitespace, so that shouldn't really affect anything.

I do see the occasional functional change, i.e. move something to a variable rather than inline, so it would be good to have a quick smoke test for that.

Good to have this in otherwise, especially in CI!

.idea/codeStyles/Project.xml
chirayudesai: Good to have to the link here though, and CI is the best part.

grote:
> I do see the occasional functional change, i.e. move something to a variable rather than inline, so it would be good to have a quick smoke test for that.

Sometimes I moved things into a variable to have better line breaking behavior or because the style CI complained. I took care however that nothing functional was changed in this MR. Also, I did test the changes and am already working on other features based on this branch.

A bigger round of testing can happen after we merge a couple of more PRs, so we can test everything in one bunch for synergy.

chirayudesai:
> > I do see the occasional functional change, i.e. move something to a variable rather than inline, so it would be good to have a quick smoke test for that.
> 
> Sometimes I moved things into a variable to have better line breaking behavior or because the style CI complained. I took care however that nothing functional was changed in this MR. Also, I did test the changes and am already working on other features based on this branch.

Agree, maybe functional was the wrong word, I just meant non-whitespace.
> 
> A bigger round of testing can happen after we merge a couple of more PRs, so we can test everything in one bunch for synergy.

Sure, a quick smoke test is fine for stuff like this, and then we can do a full testing cycle afterwards.

chirayudesai (approved):

PR #140, opened 2020-10-07
Add backup of local contacts

### LINKED ISSUE #66: Backup on-device contacts
We should also backup Contacts that are only stored on-device aka those that are not synced via a CardDAV or other sync solution.

grote:
Backup is done with a custom backup agent that does a full backup, mainly to give us more quota that might be required for contact photos. An incremental K/V backup would have been nice, but is limited to 5/10MB.

The serialization and de-serialization uses AOSP's vcard library to keep the code on our side as small as possible. The code is written in Java, not Kotlin, so if ever needed, we could integrate this into AOSP itself as a patch to some existing app maybe.

The backup was implemented as an extra APK so we can use the existing system backup API
and keep this code (and the tests as well as permissions) nicely separate from Seedvault itself.
AOSP upstream uses the same approach.

OEM integrators won't have extra work, because Seedvault's `Android.bp` adds Local Contact Backup as a dependency, so it will be pulled in automatically from the same git repo.

This MR adds unit and integration tests for the contact backup. The latter fail if you have preexisting contacts to prevent them from getting wiped unintentionally by the test.

Closes #66


contactsbackup/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Permissions
+
+* `android.permission.READ_CONTACTS` to back up local contacts.
+* `android.permission.WRITE_CONTACTS` to restore local contacts to the device.
+
+## License
+
+This application is available as Free Software
+under the terms of the [GPLv3 (or later) License](https://opensource.org/licenses/GPL-3.0).
```
chirayudesai, line 15: This will just make things complicated, Apache 2 is ideal for code that's meant for use in AOSP, like Seedvault.

chirayudesai (changes requested):

contactsbackup/README.md
stevesoltys: I agree with @chirayudesai. I think it's better to have the more permissive license.

chirayudesai (dismissed):
Code looks good, nice and simple.

Untested personally, but we can do a nice backup run with these recent changes and then cut a stable 'android11' based on that.

chirayudesai (approved):
Looks like you just rebased this commit

PR #145, opened 2020-10-13
Re-install apps with APK splits if they are compatible

### LINKED ISSUE #91: Apps using app bundles crash after getting restored
Some apps do get installed on a restore, but they crash immediately when you try to launch them.

Need to uninstall and reinstall to get them working.

I've seen this with Twitter, and then some other app prompted me to something about needing Google Play to get the needed updates. This was also faced by somebody on #CalyxOS:matrix.org

I think this could be related to split APKs, need to check.

I do have a backup that can reproduce this, will get logs.

grote:
This is the second part of #144 and builds upon this work to properly re-install apps even if they have APK splits. It introduces a compatibility checker which only looks at screen density and cpu architecture for now. This would work for all the splits I've seen so far. Unusual splits can be reported with https://gitlab.com/grote/app-bundle-reporter

More details as always can be found in the individual commit messages.

Fixes #91

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
@@ -1,0 +1,118 @@
+package com.stevesoltys.seedvault.restore.install
+
+import android.content.Context
+import android.os.Build
+import android.util.Log
+
+private const val TAG = "SplitCompatChecker"
+
+private const val CONFIG_PREFIX = "config."
+private const val CONFIG_LENGTH = CONFIG_PREFIX.length
+
+private const val LDPI_VALUE = 120
+private const val MDPI_VALUE = 160
+private const val TVDPI_VALUE = 213
+private const val HDPI_VALUE = 240
+private const val XHDPI_VALUE = 320
+private const val XXHDPI_VALUE = 480
+private const val XXXHDPI_VALUE = 640
+
+class DeviceInfo(context: Context) {
+    val densityDpi: Int = context.resources.displayMetrics.densityDpi
+    val supportedABIs: List<String> = Build.SUPPORTED_ABIS.toList()
+}
+
+/**
+ * Tries to determine APK split compatibility with a device by examining the list of split names.
+ * This only looks on the supported ABIs and the screen density.
+ * Other config splits e.g. based on OpenGL or Vulkan version are also possible,
+ * but don't seem to be widely used, so we don't consider those for now.
+ */
+class ApkSplitCompatibilityChecker(private val deviceInfo: DeviceInfo) {
+
+    private val abiMap = mapOf(
+        "armeabi" to "armeabi",
+        "armeabi_v7a" to "armeabi-v7a",
+        "arm64_v8a" to "arm64-v8a",
+        "x86" to "x86",
+        "x86_64" to "x86_64",
+        "mips" to "mips",
+        "mips64" to "mips64"
+    )
+    private val densityMap = mapOf(
+        "ldpi" to LDPI_VALUE,
+        "mdpi" to MDPI_VALUE,
+        "tvdpi" to TVDPI_VALUE,
+        "hdpi" to HDPI_VALUE,
+        "xhdpi" to XHDPI_VALUE,
+        "xxhdpi" to XXHDPI_VALUE,
+        "xxxhdpi" to XXXHDPI_VALUE
+    )
+
+    /**
+     * Returns true if the list of splits can be considered compatible with the current device,
+     * and false otherwise.
+     */
+    fun isCompatible(splitNames: Collection<String>): Boolean = splitNames.all { splitName ->
+        // all individual splits need to be compatible (which can be hard to judge by name only)
+        isCompatible(splitName)
+    }
+
+    private fun isCompatible(splitName: String): Boolean {
+        val index = splitName.indexOf(CONFIG_PREFIX)
+        // If this is not a standardized config split, we just assume that it will work,
+        // as it is most likely a dynamic feature module.
+        if (index == -1) {
+            Log.v(TAG, "Not a config split '$splitName'. Assuming it is ok.")
+            return true
+        }
+
+        val name = splitName.substring(index + CONFIG_LENGTH)
+
+        // Check if this is a known ABI config
+        if (abiMap.containsKey(name)) {
+            // The ABI split must be supported by the current device
+            return isAbiCompatible(name)
+        }
+
+        // Check if this is a known screen density config
+        densityMap[name]?.let { splitDensity ->
+            // the split's density must not be much lower than the device's.
+            return isDensityCompatible(splitDensity)
+        }
+
+        // At this point we don't know what to make of that split,
+        // so let's just hope that it will work. It might just be a language.
+        Log.v(TAG, "Unhandled config split '$splitName'. Assuming it is ok.")
+        return true
+    }
+
+    private fun isAbiCompatible(name: String): Boolean {
+        return if (deviceInfo.supportedABIs.contains(abiMap[name])) {
+            Log.v(TAG, "Config split '$name' is supported ABI (${deviceInfo.supportedABIs})")
+            true
+        } else {
+            Log.w(TAG, "Config split '$name' is not supported ABI (${deviceInfo.supportedABIs})")
+            false
+        }
+    }
+
+    private fun isDensityCompatible(splitDensity: Int): Boolean {
+        @Suppress("MagicNumber")
+        val acceptableDiff = deviceInfo.densityDpi / 3
+        return if (deviceInfo.densityDpi - splitDensity > acceptableDiff) {
+            Log.w(
+                TAG,
+                "Config split density $splitDensity not compatible with ${deviceInfo.densityDpi}"
+            )
+            false
+        } else {
+            Log.v(
+                TAG,
+                "Config split density $splitDensity compatible with ${deviceInfo.densityDpi}"
+            )
+            true
+        }
+    }
+
+}
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
@@ -1,7 +1,5 @@
 package com.stevesoltys.seedvault.restore.install
 
-import android.content.Context
-import android.os.Build
 import android.util.Log
 
 private const val TAG = "SplitCompatChecker"
@@ -9,6 +7,7 @@ private const val TAG = "SplitCompatChecker"
 private const val CONFIG_PREFIX = "config."
 private const val CONFIG_LENGTH = CONFIG_PREFIX.length
 
+// see https://developer.android.com/training/multiscreen/screendensities#TaskProvideAltBmp
 private const val LDPI_VALUE = 120
 private const val MDPI_VALUE = 160
 private const val TVDPI_VALUE = 213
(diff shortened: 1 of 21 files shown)


app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
```diff
@@ -0,0 +1,118 @@
...
+import android.content.Context
+import android.os.Build
+import android.util.Log
+
+private const val TAG = "SplitCompatChecker"
+
+private const val CONFIG_PREFIX = "config."
+private const val CONFIG_LENGTH = CONFIG_PREFIX.length
+
+private const val LDPI_VALUE = 120
```
chirayudesai, line 12: You should add a link to the source here, https://github.com/google/bundletool/blob/1.2.0/src/main/java/com/android/tools/build/bundletool/model/utils/ResourcesUtils.java#L72

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
```diff
@@ -0,0 +1,118 @@
...
+class DeviceInfo(context: Context) {
+    val densityDpi: Int = context.resources.displayMetrics.densityDpi
+    val supportedABIs: List<String> = Build.SUPPORTED_ABIS.toList()
+}
+
+/**
+ * Tries to determine APK split compatibility with a device by examining the list of split names.
+ * This only looks on the supported ABIs and the screen density.
+ * Other config splits e.g. based on OpenGL or Vulkan version are also possible,
+ * but don't seem to be widely used, so we don't consider those for now.
```
chirayudesai, line 29: I feel like we've only gone through a small subset of apps, I don't remember seeing any newer games on the list which would be most likely to utilize such a split.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
```diff
@@ -0,0 +1,118 @@
...
+     * and false otherwise.
+     */
+    fun isCompatible(splitNames: Collection<String>): Boolean = splitNames.all { splitName ->
+        // all individual splits need to be compatible (which can be hard to judge by name only)
+        isCompatible(splitName)
+    }
+
+    private fun isCompatible(splitName: String): Boolean {
+        val index = splitName.indexOf(CONFIG_PREFIX)
+        // If this is not a standardized config split, we just assume that it will work,
```
chirayudesai, line 63: I feel like this is the wrong assumption to make here.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        // Check if this is a known screen density config
+        densityMap[name]?.let { splitDensity ->
+            // the split's density must not be much lower than the device's.
+            return isDensityCompatible(splitDensity)
+        }
+
+        // At this point we don't know what to make of that split,
+        // so let's just hope that it will work. It might just be a language.
```
chirayudesai, line 85: We can handle languages like abi and density too, and the bundletool class linked above has some relevant code too.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
```diff
@@ -0,0 +1,118 @@
...
+        Log.v(TAG, "Unhandled config split '$splitName'. Assuming it is ok.")
```
chirayudesai, line 86: That would then mean that we can assume otherwise, unless it's the same device model.

chirayudesai (changes requested):
I feel like we may not want to just treat any random splits as compatible all the time, but do it only when the device is the same.

Which is why I mentioned handling languages in that config, as that means we've then handled the most common of splits.

OpenGL/Vulkan wasn't seen so far, but I imagine that could be the next common target once we started seeing it as that's something generic and not app specific.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
grote: Even if I took those numbers from https://developer.android.com/training/multiscreen/screendensities#TaskProvideAltBmp ?

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
grote: I checked several games on apk mirror, but most just used arch based splits, some also languages. Please try to find more. I agree that it would be best to have a much larger data basis for this.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
grote: Possibly true, but please provide examples here where the assumption is wrong.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
grote: We could, but what would be the benefit?

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
chirayudesai: Yes. :trollface:

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
chirayudesai: Assuming something unknown just works is wrong was my point.

I don't think we have sufficient data yet to make that claim, and till since we already have a link to store solution, that'd work well.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
chirayudesai: Read the comment below, or the github comment of the full review

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
```diff
@@ -0,0 +1,134 @@
...
+private const val MDPI_VALUE = 160
+private const val TVDPI_VALUE = 213
+private const val HDPI_VALUE = 240
+private const val XHDPI_VALUE = 320
+private const val XXHDPI_VALUE = 480
+private const val XXXHDPI_VALUE = 640
+
+/**
+ * Tries to determine APK split compatibility with a device by examining the list of split names.
+ * This only looks on the supported ABIs and the screen density.
```
chirayudesai, line 21: and language now

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
```diff
@@ -0,0 +1,134 @@
...
+            return true
+        }
+
+        // At this point we don't know what to make of that split
+        return if (deviceInfo.isSameDevice(deviceName)) {
+            // so we only allow it if it came from the same device
+            Log.v(TAG, "Unhandled config split '$splitName'. Came from same device.")
+            true
+        } else {
+            Log.w(TAG, "Unhandled config split '$splitName'. Not allowed.")
```
chirayudesai, line 101: Would we want to allow these when `unknownAllowed` is true?

app/src/main/java/com/stevesoltys/seedvault/restore/install/DeviceInfo.kt
```diff
@@ -0,0 +1,27 @@
...
+import android.content.Context
+import android.os.Build
+import com.android.internal.app.LocalePicker
+import com.stevesoltys.seedvault.R
+
+class DeviceInfo(context: Context) {
+    val densityDpi: Int = context.resources.displayMetrics.densityDpi
+    val supportedABIs: List<String> = Build.SUPPORTED_ABIS.toList()
+    private val deviceName: String = "${Build.MANUFACTURER} ${Build.MODEL}"
+    private val languages = LocalePicker.getSupportedLocales(context)
```
chirayudesai, line 12: We could add a comment here about what we based this on to make future updates easier

app/src/main/res/values/strings.xml
```diff
@@ -118,6 +118,8 @@
     <string name="restore_installing_tap_to_install">Tap to install</string>
     <string name="restore_next">Next</string>
     <string name="restore_restoring">Restoring backup</string>
+    <string name="restore_restoring_error_title">Not all apps restored</string>
+    <string name="restore_restoring_error_message">You can still try to re-install these apps later with "auto restore" turned on to get their data restored.</string>
```
chirayudesai, line 122: This is where apk install worked but app data restore failed right?

We could tweak this message a bit, it's already quite long but also sounds confusing.

chirayudesai (approved):
Just some minor comments inline, looks good otherwise.

chirayudesai:
Oh and I'm yet to test this.

@grote did you try a full SetupWizard restore sequence with this yet?

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
grote: I think this is too risky as there can be lots of crazy splits and we have no way to check compatibility here.

app/src/main/java/com/stevesoltys/seedvault/restore/install/DeviceInfo.kt
grote: What do you mean?

app/src/main/res/values/strings.xml
grote: > This is where apk install worked but app data restore failed right?

This is shown after data restore when there are aps whose data could not be restored because they were not installed, but we have data in the backup for auto-restore later.

Suggestions for how to improve it are most welcome!

app/src/main/java/com/stevesoltys/seedvault/restore/install/DeviceInfo.kt
chirayudesai: Nevermind.

app/src/main/res/values/strings.xml
chirayudesai: @mikeNG thoughts?

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
chirayudesai: Ok so unknown splits are ok if enabled, unless they're unknown config.* splits, since for those we have a well-defined list of what we accept. Right?

stevesoltys (approved):

app/src/main/res/values/strings.xml
mikeNG: Does the user see which applications failed? If yes, then maybe this sounds better?

Unable to restore some apps
You can reinstall these apps manually and "Automatic restore" will attempt to restore their data.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkSplitCompatibilityChecker.kt
grote: Right.

chirayudesai (approved):

PR #153, opened 2020-10-23
Move to different versioning scheme and show version in About dialog

grote:


Android.bp
```diff
@@ -49,6 +49,8 @@ android_app {
     optimize: {
         enabled: false,
     },
+    sdk_version: "30",
```
chirayudesai, line 52: Could set to current, not sure if that's handled any differently though.

app/src/main/AndroidManifest.xml
```diff
@@ -2,8 +2,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     package="com.stevesoltys.seedvault"
-    android:versionCode="7"
-    android:versionName="1.0.0">
+    android:versionCode="11000001"
```
chirayudesai, line 5: I thought we wanted to use API for versionCode, so `30000001`

contactsbackup/Android.bp
```diff
@@ -13,7 +13,8 @@ android_app {
     required: [
         "default-permissions_org.calyxos.backup.contacts",
     ],
-    sdk_version: "current",
+    sdk_version: "30",
```
chirayudesai, line 16: Same as above

contactsbackup/src/main/AndroidManifest.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    package="org.calyxos.backup.contacts">
+    package="org.calyxos.backup.contacts"
+    android:versionCode="11000001"
```
chirayudesai, line 5: Same as above

Android.bp
grote: I set it to `30` explicitly, so when you run this on Android 12 without making the required modifications, you'll notice here.

app/src/main/AndroidManifest.xml
grote: Good point, in case we need to target a possible `31` with a different version.

grote: Fixed.

Android.bp
chirayudesai: Shouldn't it be the other way around for that.

Anyway, we shouldn't be making changes now for something next year, it's easy enough to know you're trying to build this without any changes in a new version.

And it might not even work, due to now using hidden-apis

"current" is handled differently from a specific version

chirayudesai (changes requested):

chirayudesai:
Since we're touching `Android.bp`, I think we should do a quick smoke test before merge.

As long as it launches and can start a backup, all good.

Android.bp
grote: removed

grote:
Latest version doesn't touch `Android.bp` anymore.

chirayudesai (approved):
We should set a versioning scheme.

I propose just bump early, bump often, without thinking too much. Could be a bump for PR, or even better any major features or bugfixes.

App bundles? bump
Fix nextcloud? bump
Improve usb? bump

grote:
> I propose just bump early, bump often, without thinking too much.

Agreed!

stevesoltys (approved):
Looks good. The commit message has a small typo btw :smile:

chirayudesai:
@grote gonna let you fix the typo.

Let's pick this to 10 if there aren't any major conflicts, and then we bump the version there and do a release.

@stevesoltys maybe now we can finally move this to an organisation as previously discussed :)

stevesoltys:
Sounds good! Once we're ready to do a release we'll migrate to the organization as well.

grote:
typo fixed

chirayudesai (approved):

Issue #117, opened 2020-09-08
UI Redesign to improve existing and add new features


grote:
Seedvault's UI either needs to be redesigned or the current design adapted to accommodate additional functionality we want to add to the app.

I give a feature list below to have an overview over new and existing functionality:

* Initial Setup
  * **[NEW]** Setup in devices setup-wizard on first boot #63
  * **[NEW]** Initial onboarding explaining subsequent step #86
  * recovery code writing down and re-entering for confirmation
  * backup storage selection
* Status information
   * time of last backup
   * current backup location
   * app status information
       * app backup status (ok, not allowed, no data, error, etc.)
       * time of last app backup
       * app included/excluded from backup
    * **[NEW]** size of backup? (unclear if should be shown)
    * **[NEW]** time of last SMS (texts) backup (unclear if should be shown)
    * **[NEW]** time of last call log backup (unclear if should be shown) #93
    * **[NEW]** time of last local contacts backup (unclear if should be shown) #66
    * **[NEW]** time of last settings backup (unclear if should be shown)
    * **[NEW]** time (and size?) of last external data storage backup (unclear if should be shown, per folder?) #36 #57
* Settings
  * backup on/off
  * change backup location
  * enable/disable automatic restore
  * backup off app APKs on/off
  * exclude individual apps from backup
  * **[NEW]** add/remove external folder to get backed up (e.g. Photos, Music, Signal Backup)  #36 #57
* **[NEW]** Restore Set management #100
  * **[NEW]** list all backups in the storage location (and their time and size?)
  * **[NEW]** start using a new/fresh backup for subsequent backup (leaves current/old ones available for later)
  * **[NEW]** delete old backups
* **[NEW]** verify recovery code (to ensure we have the right one)
* **[NEW]** change recovery code if lost (will delete all old backups, if possible) #62
* Make/trigger backup now
* About screen

Currently, status and settings are somewhat mixed and there has been feedback that users weren't sure if backup is working. However, this also related to the inability to do (full) restores without factory reset.

chirayudesai:
> Seedvault's UI either needs to be redesigned or the current design adapted to accommodate additional functionality we want to add to the app.
> 
> I give a feature list below to have an overview over new and existing functionality:
> 
> * Initial Setup
>   
>   * **[NEW]** Setup in devices setup-wizard  #63
"Setup from SetupWizard on first boot"
[...]

PR #154, opened 2020-10-23
Show backup status of select system apps

grote:
A nice improvement before we tackle #117.

device-2020-10-23-135659


stevesoltys (approved):

app/src/main/res/values/strings.xml
```diff
@@ -91,6 +91,10 @@
 
     <!-- App Backup and Restore State -->
 
+    <string name="backup_sms">SMS text messages</string>
```
chirayudesai, line 94: "SMS text messages" one feels somewhat redundant - and this might include MMS too.

chirayudesai (commented):
@mikeNG string review appreciated.

app/src/main/res/values/strings.xml
grote: Google uses the same language though: https://github.com/stevesoltys/seedvault/issues/117#issuecomment-701335659
Also, I think MMS might not always get included as Google's non-free backup as a special class for that.

mikeNG:
> @mikeNG string review appreciated.

Looks fine

PR #160, opened 2020-11-09
Move CI to GitHub Actions

stevesoltys:


.travis.yml
```diff
@@ -1,33 +0,0 @@
-dist: trusty
-
-jdk:
-  - openjdk11
-
-language: android
-android:
-  components:
-    - build-tools-30.0.2
-    - android-30
```
chirayudesai, line 10: Don't need to set this either?

Does this figure out the versions out automatically somehow based on gradle config?

stevesoltys: This was to tell Travis what Android version to install on the build image. I think all versions of the Android SDK are installed on the GitHub build image.

This seems to be what determines which Android version the build uses: https://github.com/seedvault-app/seedvault/blob/master/app/build.gradle#L20

chirayudesai (dismissed):
Just gotta disable "Travis CI - Pull Request Expected — Waiting for status to be reported
Required" then

.travis.yml
chirayudesai: Sweet

grote (changes requested):
We should also change/remove this:
https://github.com/seedvault-app/seedvault/blob/7694eb3cf9052732878715eb1acc89957bf8da35/app/build.gradle#L21

Also, will the travis build badge still work after this?

chirayudesai (approved):

grote (approved):

PR #163, opened 2020-11-14
Add SeedvaultOverlay RRO to reduce ASOP integretion steps and update README

phhusson:


README.md
```diff
@@ -15,10 +15,6 @@ A backup application for the [Android Open Source Project](https://source.androi
 
 For older versions of Android, check out [the branches](https://github.com/stevesoltys/seedvault/branches).
 
-## Getting Started
```
stevesoltys, line 18: Any reason for removing this? The wiki contains a lot of helpful information.

stevesoltys (changes requested):

README.md
stevesoltys: I created #164 to track the issue of it being outdated.

phhusson: Well, I think that having basic AOSP integration in README is nice, wiki isn't up-to-date, and it's redundant to have both in the README. Also the section is called "Getting started", but to me "getting started" would explain how to build it ouf of tree and install it manually. But that's really up to you. I can remove the README.md changes for the moment, and just have the overlay.

Android.bp
```diff
@@ -44,7 +44,8 @@ android_app {
     required: [
         "LocalContactsBackup",
         "privapp_whitelist_com.stevesoltys.backup",
-        "com.stevesoltys.backup_whitelist"
+        "com.stevesoltys.backup_whitelist",
+        "SeedvaultOverlay",
```
chirayudesai, line 48: This should not be added here. Not everybody that ships Seedvault would want to set it as the default, for example LineageOS still defaults to the `gms` provider and instead uses a patch to add a selector to Settings.

* overlay: https://review.lineageos.org/c/LineageOS/android_vendor_lineage/+/288544
* Settings patch: https://review.lineageos.org/c/LineageOS/android_packages_apps_Settings/+/288543

README.md
chirayudesai: I guess we can do just the README changes in a separate PR.

Wiki helps - there's more there than just the default overlay, a patch is linked to actually enable this for existing users as well (who may have this set to the default), and not just new users/installs.

README.md
```diff
@@ -43,5 +39,15 @@ The [Seedvault backup parser](https://github.com/tlambertz/seedvault_backup_pars
 allows you to decrypt and inspect your backups.
 It can also re-encrypt them.
 
+## AOSP build
+
+To build in AOSP, add
+```
+	<remote name="seedvault-github" fetch="https://github.com/seedvault-app/" />
+    <project path="vendor/seedvault" name="seedvault" remote="seedvault-github" revision="master" />
```
chirayudesai, line 47: I'd rather do `packages/apps/Seedvault`

chirayudesai (changes requested):
This is nice.

chirayudesai:
Closing since it's gone a while without any updates.

PR #166, opened 2020-11-19
Group app status list into three sections

grote:
Part of #117

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -96,14 +96,13 @@
     <string name="backup_settings">Device settings</string>
     <string name="backup_call_log">Call history</string>
     <string name="backup_contacts">Local contacts</string>
-    <string name="backup_section_user">User Apps</string>
+    <string name="backup_section_user">Installed Apps</string>
     <!-- This text gets shown for apps that the OS did not try to backup for whatever reason e.g. no backup was run yet -->
     <string name="backup_app_not_yet_backed_up">Waiting to back up…</string>
     <string name="restore_app_not_yet_backed_up">Was not yet backed up</string>
     <string name="backup_app_was_stopped">Not backed up as it wasn\'t used recently</string>
     <string name="restore_app_was_stopped">Was not backed up as it hadn\'t been used recently</string>
     <string name="backup_app_no_data">App reported no data for backup</string>
-    <string name="backup_app_not_allowed">App doesn\'t allow backup</string> <!-- TODO remove -->
     <string name="restore_app_not_allowed">App didn\'t allow backup</string>
     <string name="backup_app_quota_exceeded">Backup quota exceeded</string>
     <string name="restore_app_quota_exceeded">Backup quota was exceeded</string>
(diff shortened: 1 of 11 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -91,21 +91,24 @@
...
     <string name="backup_contacts">Local contacts</string>
+    <string name="backup_section_user">User Apps</string>
     <!-- This text gets shown for apps that the OS did not try to backup for whatever reason e.g. no backup was run yet -->
     <string name="backup_app_not_yet_backed_up">Waiting to back up…</string>
     <string name="restore_app_not_yet_backed_up">Was not yet backed up</string>
     <string name="backup_app_was_stopped">Not backed up as it wasn\'t used recently</string>
     <string name="restore_app_was_stopped">Was not backed up as it hadn\'t been used recently</string>
     <string name="backup_app_no_data">App reported no data for backup</string>
-    <string name="backup_app_not_allowed">App doesn\'t allow backup</string>
+    <string name="backup_app_not_allowed">App doesn\'t allow backup</string> <!-- TODO remove -->
```
stevesoltys, line 106: Just curious, anything blocking now for removing this? And why remove it?

stevesoltys (approved):

app/src/main/res/values/strings.xml
grote: Good catch! As the not allowed apps are now in their own section without repeating their status text, this string isn't needed anymore. I've now removed it.

stevesoltys (approved):

app/src/main/res/values/strings.xml
```diff
@@ -91,21 +91,24 @@
 
     <!-- App Backup and Restore State -->
 
+    <string name="backup_section_system">System Apps</string>
     <string name="backup_sms">SMS text messages</string>
     <string name="backup_settings">Device settings</string>
     <string name="backup_call_log">Call history</string>
     <string name="backup_contacts">Local contacts</string>
+    <string name="backup_section_user">User Apps</string>
```
chirayudesai, line 99: How about "Installed Apps" instead?

app/src/main/res/values/strings.xml
```diff
@@ -91,21 +91,24 @@
...
     <string name="backup_app_was_stopped">Not backed up as it wasn\'t used recently</string>
     <string name="restore_app_was_stopped">Was not backed up as it hadn\'t been used recently</string>
     <string name="backup_app_no_data">App reported no data for backup</string>
-    <string name="backup_app_not_allowed">App doesn\'t allow backup</string>
+    <string name="backup_app_not_allowed">App doesn\'t allow backup</string> <!-- TODO remove -->
     <string name="restore_app_not_allowed">App didn\'t allow backup</string>
     <string name="backup_app_quota_exceeded">Backup quota exceeded</string>
     <string name="restore_app_quota_exceeded">Backup quota was exceeded</string>
     <string name="restore_app_not_installed">App not installed</string>
+    <string name="backup_section_not_allowed">Apps that do not allow data backup</string>
```
chirayudesai, line 110: Could just call it "do not allow backup".

That can be confusing with APKs though.

grote: yeah I think we have the space and the additional word might provide more clarity

app/src/main/res/values/strings.xml
grote: would be fine with me as well. I used what the design had.

chirayudesai: "Installed" is just what's used everywhere in the OS I think, let's make the change and then this can be merged.

grote: done

chirayudesai (approved):

PR #191, opened 2021-01-11
Remove redirect from README

williamtheaker:
Point towards the new repo URI to prevent confusion so contributors don't think this repo is a downstream fork.


grote (approved):
Thanks for catching this!

PR #220, opened 2021-03-04
Change signing configuration to use AOSP platform key

chirayudesai:
* https://stackoverflow.com/a/40145242

I've been wanting to do this since a while. This will make it easier to test CI built apps or just Studio built apps with the regular OS debug builds. OS Release builds would be signed with a different key usually (more like re-signed but w/e)

Change-Id: I12d09f299f4f0f7d8879a631d4f3298a18409683


chirayudesai:
Could still keep the current config for release builds if desired.

grote (approved):
Nice, great work!

Also works with my local setup which I can now retire.

PR #221, opened 2021-03-05
Add storage backup feature

### LINKED ISSUE #57: Backing up the internal storage - /sdcard
We should backup the entirety of the internal storage as well.
While it might be too big for certain cloud providers, it would be just fine if backing up to removable storage.

It also has the added benefit of backing up what we want from #36 , i.e. backing up internal storage would also backup WhatsApp's directory which would include it's chat backup databases and media files as well.

grote:
Closes #57


app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
```diff
@@ -126,6 +126,13 @@ internal class SettingsViewModel(
             networkCallback.registered = true
         }
 
+        if (settingsManager.isStorageBackupEnabled()) {
+            // disable storage backup if new storage is on USB
```
chirayudesai, line 132: Why?

chirayudesai (commented):
Went through everything but "Add storage library (and demo app)" - seems ok.

I'll go through that later.

One thought:
Why not move the storage backup to it's own subdir - `.SeedVaultAndroidBackup/storage` could work.

I was trying to browse the main folder `DIRECTORY_ROOT` and the 256 folders made it really slow to open that folder.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
grote: Because the chance that you have your usb device plugged in what the job runs is quite low. We start the job when you insert the flash drive, so it shouldn't start automatically.

>  Why not move the storage backup to it's own subdir - .SeedVaultAndroidBackup/storage could work.

Every sub-folder will make access through SAF slower, so as little subfolders as needed. Also, maybe later we want to store app data or APKs in this scheme as well to better hide what apps you use.

chirayudesai: > Because the chance that you have your usb device plugged in what the job runs is quite low. We start the job when you insert the flash drive, so it shouldn't start automatically.
> 
What I did was:
* Location internal storage, enabled storage backup, selected some files.
* Backup now, check `.SeedVaultAndroidBackup` - saw the 256 folders after a momentary delay..
* Change location to USB, backup now.
* Checked `.SeedVaultAndroidBackup` on usb - no 256 folders.

> > Why not move the storage backup to it's own subdir - .SeedVaultAndroidBackup/storage could work.
> 
> Every sub-folder will make access through SAF slower, so as little subfolders as needed. Also, maybe later we want to store app data or APKs in this scheme as well to better hide what apps you use.

Far enough.

chirayudesai: Logcat i see around 3 - https://zifb.in/A1DmN7Y7Nx

If I unplug and re-plug the USB drive, and then "Backup now" again, it seems to create those folders and do a full storage backup.

grote: > Checked .SeedVaultAndroidBackup on usb - no 256 folders.

yeah the folders get created when the job actually runs

> I would suggest .SeedVaultStorageBackup but maybe we don't want to do that for other reasons.

yeah this would be a different root/storage location we'd need permission to store things in again.

grote: > Logcat i see around 3 - https://zifb.in/A1DmN7Y7Nx

This crash happens where a job gets scheduled in case the storage that is being changed to is *not* USB. So I wonder why you got this after changing to USB storage. Is it possible that it happened earlier? Is it possible that you did not wait for the backup to complete before changing the location?

I don't understand why it would say

> ` No such service ComponentInfo{com.stevesoltys.seedvault/com.stevesoltys.seedvault.storage.StorageBackupJobService}`

when this component clearly exists. Can you reproduce the problem? I tried with your instructions and failed.

chirayudesai: > > Logcat i see around 3 - https://zifb.in/A1DmN7Y7Nx
> 
> This crash happens where a job gets scheduled in case the storage that is being changed to is _not_ USB. So I wonder why you got this after changing to USB storage. Is it possible that it happened earlier? Is it possible that you did not wait for the backup to complete before changing the location?
> 
I definitely waited for it to complete, there weren't many files anyway.

> I don't understand why it would say
> 
> > ` No such service ComponentInfo{com.stevesoltys.seedvault/com.stevesoltys.seedvault.storage.StorageBackupJobService}`
> 
> when this component clearly exists. Can you reproduce the problem? I tried with your instructions and failed.

I think I tried it twice that day when I first reported it, but I'll see if I can reproduce on another device.

chirayudesai:
Rebased so that CI can run.

chirayudesai (approved):
The storage library has been extensively tested over the last few months, and this patch for the last few weeks.

Works really well, we found some issues and fixed them already.

It's still marked as experimental so I think the occasional issue is fine, and we definitely need all the additional testing we can get.

PR #227, opened 2021-03-22
Fix pluralization in german strings

bananer:


chirayudesai:
We handle translations through weblate, would you mind making this edit there?

It's at https://hosted.weblate.org/projects/calyxos/seedvault/

Not merging this here to keep translations in one place and avoid conflicts.

PR #236, opened 2021-04-09
Improve recovery code output fragment's wordlist layout

luk1337:
* Enable scrollbar
* Enable text size autoscalling (min 12sp, max 24sp)
* Fix wordList constraints
* Set wordList height to match_parent so that scrollbar
  our word list will always start from top
* Set wordList width to match_parent so that scrollbar
  is displayed on the sides
* Reduce landscape spanCount to 3 so that we'll be less
  likely to clip the some longer word

Change-Id: I7a2481fe25163f81f6d3c3626784cc4681f7c4ef

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/recoverycode/RecoveryCodeOutputFragment.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/recoverycode/RecoveryCodeOutputFragment.kt
@@ -45,7 +45,7 @@ class RecoveryCodeOutputFragment : Fragment() {
     private fun setGridParameters(list: RecyclerView) {
         val layoutManager = list.layoutManager as GridLayoutManager
         if (resources.configuration.orientation == ORIENTATION_LANDSCAPE) {
-            layoutManager.spanCount = 4
+            layoutManager.spanCount = 3
         } else {
             layoutManager.spanCount = 2
         }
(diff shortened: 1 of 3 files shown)


luk1337:
Mostly affects 16:9 devices with display size / font-size set to high values.

grote (approved):
I didn't run this code, but assume @luk1337 did that. The changes look ok.

luk1337:
@grote yeah, it'd be hard to do that kind of stuff blind.

chirayudesai:
This makes it look really small on bigger devices now.

Screenshot_20210416-232344

That's bramble, 1080x2340, 420dpi. It would fill the whole screen nicely before, now the text is way too small.

chirayudesai (changes requested):
Should ideally stay as close to before as possible on other devices.

luk1337:
Try removing android:autoSize{Min,Max}TextSize and testing it on different DPIs.

chirayudesai:
> Try removing android:autoSize{Min,Max}TextSize and testing it on different DPIs. Everything else should be ok

Now it looks ok, but cuts off words on larger/largest densities.

chirayudesai (approved):
Works much better, thanks!

grote (changes requested):
To be honest, I am not happy with the hardcoding of the list item height and some other changes. I pushed an alternative solution to https://github.com/grote/seedvault/tree/recovery-code-output-small-screens

chirayudesai:
Merged https://github.com/seedvault-app/seedvault/pull/276 instead.

Makes it look slightly different on all devices, and a lot better on smaller screens / larger dpis.

image

This has a scrollbar, compared to cutoff text and no scrolling before.

PR #238, opened 2021-04-12
Also consider contacts with account 'com.android.contacts' for backup

### LINKED ISSUE #237: Message "Waiting to back up..." for local contacts
I am using seedvault on lineageos 18.1.  The "Local contacts" entry is stuck at "Waiting to back up...", even though I have contacts on device (I don't have any sort of cloud syncing enabled for them).  Is there something I need to enable to get this to work?

grote:
In our tests with CalyxOS, all local contacts had a `null` account, but for some reason on LineageOS, those contacts have `com.android.contacts`. So we now consider this as well.

Fixes #237

### DIFF
--- a/contactsbackup/src/main/java/org/calyxos/backup/contacts/VCardExporter.java
+++ b/contactsbackup/src/main/java/org/calyxos/backup/contacts/VCardExporter.java
@@ -45,7 +45,7 @@ class VCardExporter {
     private Collection<String> getLookupKeys() {
         String[] projection = new String[]{LOOKUP_KEY};
         // We can not add IS_PRIMARY here as this gets lost on restored contacts
-        String selection = ACCOUNT_TYPE + " is null";
+        String selection = ACCOUNT_TYPE + " is null OR " + ACCOUNT_TYPE + "='com.android.contacts'";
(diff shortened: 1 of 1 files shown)


mikeNG (dismissed):
LineageOS indeed uses a different way to store local contacts as you can see here: https://review.lineageos.org/q/topic:%2522eleven-local-contacts%2522+status:merged

grote:
Ah good to know. I wasn't aware that local device contacts need a patch to work at all. I guess CalyxOS uses a different patch using a `null` account while LineageOS uses the contact's app package name.

mikeNG:
> Ah good to know. I wasn't aware that local device contacts need a patch to work at all. I guess CalyxOS uses a different patch using a `null` account while LineageOS uses the contact's app package name.

At this point we are just picking those patches to support upgrades from older versions that had the patches, I'm pretty sure AOSP local contacts work without any patches now.

chirayudesai (dismissed):

chirayudesai (approved):
Rebased

sam-tim1:
How do I implement this change? I'm having quite the time as a new CalyxOS user trying to fix my issue with local contacts being stuck on "waiting to back up" for weeks.

PR #241, opened 2021-04-19
Switch all text references to github.com/seedvault-app

mikeNG:


app/src/main/res/values/strings.xml
```diff
@@ -154,6 +154,6 @@
     <string name="about_author">Written by: <a href="https://github.com/stevesoltys">Steve Soltys</a> and <a href="https://blog.grobox.de">Torsten Grote</a></string>
     <string name="about_design">Design by: <a href="https://www.glennsorrentino.com/">Glenn Sorrentino</a></string>
     <string name="about_sponsor">Sponsored by: <a href="https://www.calyxinstitute.org">Calyx Institute</a> for use in <a href="https://calyxos.org">CalyxOS</a></string>
-    <string name="about_source_code">Source Code: https://github.com/stevesoltys/seedvault</string>
+    <string name="about_source_code">Source Code: https://github.com/seedvault-app/seedvault</string>
```
chirayudesai, line 157: Other languages via weblate?

Maybe in the future we could change to Source code: %1$s and put the URL in it's own non-translatable string.

Overkill now, but could be useful later.

chirayudesai (approved):

Issue #224, opened 2021-03-11
Backup Mobile Data and WiFi Permissions


bettyvschmartz:
Because there is no bulk management of network access / firewall controls, I usually end up going through apps one-by-one and blocking all network access. From there I start allowing what I need. This is an inconvenience but is really only required at first set up. 

However, these settings are not backed up so on restore one must do it all over again. 

In the absence of global toggles it would be good to have these backed up. 

-- BVS

grote:
@chirayudesai might know better, but this *sounds* like your Android ROM should include these settings in the respective app, probably settings app. This is not really a Seedvault issue, but of the app these settings come from.

chirayudesai:
@Uldiniad was going to look into for LineageOS / CalyxOS / others using the same policy-based firewall rules.

Our current idea is that there's something storing this in the system to restore the Settings after a reboot so we'll try to back that up if possible. 

The application handling the toggles (Settings or Datura) does not keep any state currently, but if the above idea doesn't work we can change that instead then to back it up there.

bettyvschmartz:
But all my other settings from the Settings are restored, it's only the permissions that aren't. Maybe the addition of the new Datura app fixes this. 

-- BVS

bettyvschmartz:
> @Uldiniad was going to look into for LineageOS / CalyxOS / others using the same policy-based firewall rules.
> 
> Our current idea is that there's something storing this in the system to restore the Settings after a reboot so we'll try to back that up if possible. 
> 
> The application handling the toggles (Settings or Datura) does not keep any state currently, but if the above idea doesn't work we can change that instead then to back it up there.

OK, this makes sense now. Would be really useful so that an app isn't restored that then gets full network access, for example. 

-- BVS

Uldiniad:
the system reads and writes to the following file `/data/system/netpolicy.xml`

grote:
I guess this is owned by the system user, so maybe some system user app could include it in its backup? I wonder why AOSP doesn't handle this. Or is the addition of this file part of a non-upstream patch?

Btw. I consider this a non-Seedvault issue, but we can keep this ticket open, if it makes sense to discuss this here.

chirayudesai:
This file does exist in AOSP. We'll look into what can access it and try to back it up accordingly - maybe in Settings or Datura or something else.

And yeah, this isn't a Seedvault issue, we can discuss it elsewhere but keep this open till this is implemented (just so we have something that users can see when reporting)

[...]

Issue #127, opened 2020-09-11
Backup format


chirayudesai:
Discussion on the format of the backup data, and future improvements.

Current format / filesystem layout:
* .SeedVaultAndroidBackup
  * $timestamp
      * $package.apk
      * $package2.apk
      * full
        * $package (file)
        * $package2 (file)
      * kv
        * $package
          * $blob
          * $blob2
          * $blob3
          * $blob4
        * $package2
          * $blob
          * $blob2
  * $timestamp2
...

The blobs listed above for kv are typically small in size, but big in number.
This sometimes leads to the case where we are working with lots and lots of small files, and that becomes inefficient fast all around.

We can try to think about a different format for the future, especially for kv.

chirayudesai:
I was exasperated by https://github.com/stevesoltys/seedvault/pull/113#issuecomment-691219295 which is what led to starting this discussion with @grote

To re-cap, it took 90 minutes to backup call logs, in the form 2000 300 byte blobs.

Surely something like a database would work better.

There can be certain tradeoffs, but with the issues we've seen also with Nextcloud and SAF, I think it would totally be worth moving off of small files in whatever way possible.

grote:
I did a [small investigation](https://github.com/grote/seedvault/commits/kv-timing) into where the time in K/V backups is spent and it isn't the crypto, as I expected at first.

Here's two example records of the initial `@pm@` backup (the rest looks similar):

```
I: 63/209 - 30% - @pm@ record for com.android.wifi.resources
V: Getting OutputStream for @pm@ com.android.wifi.resources took 1975ms
V: Writing version for @pm@ com.android.wifi.resources took 0ms
V: Encrypting header of @pm@ com.android.wifi.resources took 41ms
V: Writing header of @pm@ com.android.wifi.resources took 33ms
V: Flushing of @pm@ com.android.wifi.resources took 0ms
V: Storing @pm@ com.android.wifi.resources took 2059ms.

I: 64/209 - 31% - @pm@ record for com.android.calllogbackup
V: Getting OutputStream for @pm@ com.android.calllogbackup took 2309ms
V: Writing version for @pm@ com.android.calllogbackup took 0ms
V: Encrypting header of @pm@ com.android.calllogbackup took 25ms
V: Writing header of @pm@ com.android.calllogbackup took 27ms
V: Flushing of @pm@ com.android.calllogbackup took 0ms
V: Storing @pm@ com.android.calllogbackup took 2370ms.
```

[...]

PR #249, opened 2021-05-07
Disable AOSP K/V Call log backups

Uldiniad:
K/V Backup of call logs creates one pair per call log entry, thus
leading to thousands of files for people with lots of logs.
This is really slow especially with Nextcloud, making the backup take hours.
Fix that by not enabling the AOSP K/V code, and instead using our own
full backup code.

Full Backup implementation: https://review.calyxos.org/c/CalyxOS/platform_packages_providers_ContactsProvider/+/2970

Issue: calyxos#395
Issue: https://github.com/seedvault-app/seedvault/issues/127
Issue: https://github.com/seedvault-app/seedvault/issues/224
Change-Id: I742408ab62e126129966f7c9d095d6e61fdc58f6


chirayudesai:
@grote one thing I noticed with this change is, if you have an old backup with the K/V backup, and then upgrade to a build with this which makes a full backup - restore still restores both backups.

I was thinking of disabling `allowBackup` for the AOSP CallLogProvider, would that help with restore too?

If not, maybe we need to add some code here.

chirayudesai:
Won't be needed after https://github.com/seedvault-app/seedvault/issues/127#issuecomment-920820699

PR #262, opened 2021-06-11
Move Seedvault/LocalContactsBackup to product partition

mikeNG:
Change-Id: I0f463d4310647765266b948276d62171a7f46243


chirayudesai:
https://source.android.com/devices/bootloader/partitions/product-partitions

Product is the right partition for apps like this. Play Services lives on product on stock for example.

grote:
Did you test if something behaves differently when these apps are on `/product` and how an existing install behaves when the app location changes?

PR #273, opened 2021-07-02
Remove GPLv3: Replace BIP39 dependency with kotlin-bip39 which is MIT

### LINKED ISSUE #272: Seedvault is unintentionally implicitly GPLv3 due to NovaCrypto/BIP39 library
https://github.com/NovaCrypto/BIP39/blob/master/LICENCE.txt

We can ask its author to also license under Apache2 or move to a different library.

Since this is a security critical library, it was also suggested to implement the few bits we need ourselves to prevent upstream from making a malicious change that allows them to decrypt people's backups.

grote:
A draft only because based on #221 which changes quite a few dependency specifications.

Fixes #272


chirayudesai:
AOSP build works.

App launches, doesn't crash.

Verified code created using previous version / old library, works. Wrong code does not verify.

chirayudesai:
I wonder if it makes sense to fully vendor the lib, source code, instead of using a prebuilt for the gradle build too.

Ala storagebackup

chirayudesai:
Rebased (on the rebase of 221) so that CI can run.

chirayudesai:
`sha256sum libs/kotlin-bip39-1.0.2.jar`
* 27807752178f7a287dbb9465467d75e83092fe3509adf3bc0851c124807430ff  libs/kotlin-bip39-1.0.2.jar

Matches https://search.maven.org/artifact/cash.z.ecc.android/kotlin-bip39/1.0.2/jar

chirayudesai (approved):

PR #279, opened 2021-07-05
SeedVault 2.0: Now with storage backup!

chirayudesai:
Change-Id: Ifd639c3c9205b417d1ec37c8e7a11a2119b5cb48


chirayudesai:
@grote I noticed both storage/lib and storage/demo had version codes, would we want to bump those too for consistency?

grote:
We could do a new release of the demo app, otherwise, this is now the first release of the lib, so version code 1 is fine for now.

PR #283, opened 2021-07-06
SeedVault: Show in launcher, as "Backup"

chirayudesai:
* Without this, you have to go to Settings -> System -> Advanced - Backup
* That can be a bit too deep sometimes, and those who don't know about this
  may not find it.
* Let's add a Launcher icon so that this is easier to find.
* We call it Backup because people may not know the same SeedVault and thus
  get confused.
* Those who know the name SeedVault probably already know to look for it in
  Settings.

Change-Id: I0d21ed2a9f77786ffab482924b7c1b2cbbf050a9

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -73,7 +73,13 @@
         <activity
             android:name=".settings.SettingsActivity"
             android:permission="com.stevesoltys.seedvault.OPEN_SETTINGS"
-            android:exported="true" />
+            android:exported="true"
+            android:label="@string/backup">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+        </activity>
 
         <activity
             android:name=".ui.storage.StorageActivity"


chirayudesai:
Works fine when SeedVault is enabled.

Need to see what happens when it's not, aka when it isn't the backup transport.

app/src/main/AndroidManifest.xml
```diff
@@ -61,6 +61,15 @@
             android:name=".settings.SettingsActivity"
             android:exported="true" />
 
+        <activity-alias android:name=".settings.SettingsLink"
+                android:targetActivity=".settings.SettingsActivity"
+                android:label="@string/backup">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
```
grote, line 70: Why does it need an alias? Could the intent filter not be added to the activity above?

chirayudesai: Yeah, good point, it could be. I was initially looking at something else which had a different label, but here there's no label so it's fine.

julianfoad:
TL;DR: In my opinion it is probably best NOT to add a launcher icon by default, because that is not in the best interest of the ordinary user in their ordinary every-day activities.

The rest of this comment is rather long. It is not meant to be a rant, just a fuller explanation of why I make this suggestion, and I thought I might as well write it all down as it may not all be obvious to everyone.

I understand, it would indeed be nice to make the Seedvault UI a bit easier to find and access. It would especially be nice during the time I (the user) am setting up and testing the backup procedure, when I need to access it again and again. Let's come back to this later, towards the end of this long comment. But I am not the ordinary user and this is not the usual way of using my phone, so let's first think about the ordinary user in their ordinary usage.

I have had the privilege of watching a real Ordinary User recently and this is what I found in real life. The "ordinary user" I speak of is quite surprising to us techies: it is a person who is focused on their own activities and does not care to spend time understanding or interfering with how their phone is working. In fact we might consider them to be neglecting their phone's needs and health. This ordinary user does not want to customize their phone: they do not even add favourite apps to their home screen! They just find them in the launcher or in recents. This ordinary user does not care about keeping the software up to date, they only do so if a pop-up forces them to, otherwise they ignore it. This ordinary user does not even bother to dismiss their notifications! That's an optional extra thing they don't need to do, they don't want to waste their brain space on it, they just want to do something particular with some particular app right now and then shut the phone.

A typical default launcher such as on LineageOS shows all the user apps, plus just one "Settings" icon. That seems an intentional and user-focused way to organize things. The ordinary user likes to get some help setting up the phone, and then forget about it. They don't need or want to visit what they regard as low-level settings like these again, so they don't need easy access to them. Additional icons added to the launcher are just clutter that get in the way of them finding their user apps.

I noticed that LineageOS-for-microG adds a second settings icon for microG settings. In a small way, that is already prioritising the interests of the developers while starting to degrade experience for the ordinary user. Let us not copy that mistake.

Backup, and microG-settings, are far from the only system settings that some users might sometimes like shortcuts for. They are just two of many. It should certainly be made possible for a ROM maker or a power user to create shortcuts to settings and settings-related apps such as Seedvault, and also to particular activities within it, so that such developers or power users can make a page of settings shortcuts if they want to. That would make sense, for example, with a launcher that allows creating sub-groups of launcher icons. But while making such things possible, let us not underestimate the importance of getting the default experience right for ordinary users.

Now back to the topic of making Seedvault easier to access for those users and those times when we do need it. A few minor enhancement suggestions here; maybe we can think of more.

* Enhancement: I noticed that searching the system settings for "backup" finds Seedvault's UI, while "seedvault" does not find it. Conversely, searching for "seedvault" does find Seedvault's "App info" screen, whereas searching for "backup" does not. Let's at least add "seedvault" to the searchable string used for its settings UI, and add "backup" to the searchable string used by its "App info", to make these consistent and more discoverable.

* Enhancement: when the "Seedvault: backup running" notification is showing, clicking this should open the Seedvault UI. It currently doesn't do anything when I touch it.

* Minor enhancement of interest to power users: without adding this to the launcher, just to mark the UI activity as launchable, so that an "Open" button appears on its "app info" screen where currently there are only the Uninstall and Force Stop buttons.

(My versions: Seedvault 11-1.2, on LineageOS-for-microG 18.1 3-September-2021, on a OnePlus 6 "enchilada")

chirayudesai:
> TL;DR: In my opinion it is probably best NOT to add a launcher icon by default, because that is not in the best interest of the ordinary user in their ordinary every-day activities. Basically, my argument is that an Ordinary User will treat it as set-and-forget settings, and will want to ignore it for nearly all of their life, so we shouldn't insert an icon for it among their user apps.

The reason for having an icon is because currently it's hidden in Settings, and not all users seem to know this functionality is available. That is also the reason it's called Backup, since that is a generic term.

> 
> At first I included a long explanation here. In the interests of both being concise here and owning my own writings, I have now moved the long explanation to [my blog post](https://blog.foad.me.uk/2021/09/07/seedvault-backup-why-not-add-a-launcher-icon/) and left just the TL;DR and the concrete suggestions here.
> 
> Now on the topic of making Seedvault easier to access for those users and those times when that's needed, a few minor suggestions:
> 
> * Enhancement: I noticed that searching the system settings for "backup" finds Seedvault's UI, while "seedvault" does not find it. Conversely, searching for "seedvault" does find Seedvault's "App info" screen, whereas searching for "backup" does not. Let's at least add "seedvault" to the searchable string used for its settings UI, and add "backup" to the searchable string used by its "App info", to make these consistent and more discoverable.

This should be fixed.
> * Enhancement: when the "Seedvault: backup running" notification is showing, clicking this should open the Seedvault UI. It currently doesn't do anything when I touch it.

Same.
> * Minor enhancement of interest to power users: without adding this to the launcher, just to mark the UI activity as launchable, so that an "Open" button appears on its "app info" screen where currently there are only the Uninstall and Force Stop buttons.

Doesn't hurt, but having it available from "App info" doesn't really help that much.
> 
> Maybe we can think of more ways to make it findable when it's needed without showing anything when it's not needed.

Suggestions welcome. I still do think that an icon is good to make it easily discoverable for the user.

I believe something similar can be said for microG Settings which you mentioned on your blog - it's good to have that icon to make it easier to access, even though for that most users may just set it up once and keep using it like that afterwards.

chirayudesai:
Sadly this won't work with https://github.com/seedvault-app/seedvault/pull/287

@flawedworld thoughts?

I don't see a way to let Launcher launch it while not letting anything else.

System launcher could be given the permission but then third party apps wouldn't be able to open it.

ASerbinski:
Some applications have the ability to show and hide their launcher icon, I understand it to be through enabling/disabling the launch activity/component.

I would suggest setting it up so that it shows by default and until it has been configured for use. First launch include a big alert dialog that explains where to find it to the user, and disappears it.

chirayudesai:
Closing this here for now.

PR #286, opened 2021-07-15
Add expert settings with an option for unlimited quota

grote:
I tested this by pushing extra files to an app's private dir. It didn't get backed up due to exceeding quota with default settings, but did get backed up enabling the new expert setting.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -42,6 +42,10 @@
     <string name="settings_backup_storage_code_dialog_message">To enable storage backup, you need to first verify your recovery code or generate a new one.</string>
     <string name="settings_backup_storage_code_dialog_ok">Verify code</string>
 
+    <string name="settings_expert_title">Expert settings</string>
+    <string name="settings_expert_quota_title">Unlimited app quota</string>
+    <string name="settings_expert_quota_summary">Do not impose a limitation on the size of app backups.\n\nWarning: This can fill up your storage location quickly. Not needed for most apps.</string>
+
     <!-- Storage Location -->
     <string name="storage_fragment_backup_title">Choose where to store backups</string>
     <string name="storage_fragment_restore_title">Where to find your backups?</string>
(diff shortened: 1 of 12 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -42,6 +42,10 @@
     <string name="settings_backup_storage_code_dialog_message">To enable storage backup, you need to first verify your recovery code or generate a new one.</string>
     <string name="settings_backup_storage_code_dialog_ok">Verify code</string>
 
+    <string name="settings_expert_title">Expert settings</string>
```
chirayudesai, line 45: Maybe call it "Advanced". I'm fine either way.

app/src/main/res/values/strings.xml
```diff
@@ -42,6 +42,10 @@
...
+    <string name="settings_expert_quota_title">Unlimited app quota</string>
+    <string name="settings_expert_quota_summary">Does not impose a limitation on the size of app backups.\n\nWarning: This can fill up your storage location quickly and should not be needed. Apps that exceed their quota have a bug that should get fixed.</string>
```
chirayudesai, line 47: Made some minor changes:

Do not impose a limitation of the size of app backups.\n\nWarning: This call fill up your storage location quickly. It should not be needed in most cases, and apps that exceed their quota usually have a bug that should get fixed.

chirayudesai (changes requested):
Worked for me with one app, but not another.

Just testing by pushing big random files to said apps.

app/src/main/res/values/strings.xml
grote: I tended to expert to make sure this setting is for people that know what they are doing and are ready to bear the consequences instead of just more advanced settings.

app/src/main/res/values/strings.xml
grote: `impose a limitation on the size` sounds more correct to me, but I am not a native speaker

Using the `, and apps that ` makes the sentence longer. Usually better for readability to have more shorter sentences than few long ones.

chirayudesai: > `impose a limitation on the size` sounds more correct to me, but I am not a native speaker
> 
Ah yes I did mean 'on', that was a 2AM typo. The main change there was Does -> Do

> Using the `, and apps that ` makes the sentence longer. Usually better for readability to have more shorter sentences than few long ones.

Fair.

"Do not impose a limitation on the size of app backups.\n\nWarning: This can fill up your storage location quickly. Not needed for most apps."

Left out the part about bugs.

chirayudesai (approved):

PR #287, opened 2021-07-18
Restrict exported components

flawedworld:


grote:
@flawedworld thanks for the PR!

Do you mind adding a bit of explanation for the benefit of restricting opening those activities? Maybe also compared to the downsides of not being able to open restore activity with adb anymore?

flawedworld:
Exported components should pretty much always be restricted. It's a best practice (see CWE-926). Other applications not within the scope of ```system|signature``` should never be able to open those activities because they do not need to. Using ADB to open activities should never be done outside of a debug environment.

app/src/main/AndroidManifest.xml
```diff
@@ -47,6 +47,16 @@
         android:name="android.permission.QUERY_ALL_PACKAGES"
         tools:ignore="QueryAllPackagesPermission" />
 
+    <!-- Permission used to open settings -->
+    <permission 
+        android:name="com.stevesoltys.seedvault.opensettings"
```
mikeNG, line 52: I would suggest changing this to "com.stevesoltys.seedvault.OPEN_SETTINGS", following AOSP permission naming

app/src/main/AndroidManifest.xml
```diff
@@ -47,6 +47,16 @@
...
+        android:protectionLevel="system|signature" />
+
+    <!-- Permission used to open backup gui -->
+    <permission 
+        android:name="com.stevesoltys.seedvault.restorebackup"
```
mikeNG, line 57: Same for this, "com.stevesoltys.seedvault.RESTORE_BACKUP"

mikeNG (changes requested):

chirayudesai:
Thanks for the PR.

By default, the 'restore option' in menu isn't enabled since restoring backups to an existing system may cause some weirdness.

adb was one way around that, but not ideal.

We should add some other way to getting to the restore menu on-device, a `*#*#RESTORE#*#*` `SECRET_CODE` would work.

Although that can be done separate from this PR.

chirayudesai:
Rolled into https://github.com/seedvault-app/seedvault/pull/319 after resolving merge conflict.

PR #302, opened 2021-08-25
Disable Nextcloud restore when not installed and no store available

mikeNG:
Attempting to install Nextcloud in SetupWizard on a device
that has no app store installed results in a crash.
Disable the option and use an appropriate summary.

Change-Id: Ie3dff8d85ed68b27858fa37b3efb97eb345fa372

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootFetcher.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootFetcher.kt
@@ -156,11 +156,19 @@ internal class StorageRootFetcher(private val context: Context, private val isRe
             data = Uri.parse("nc://login/server:")
             putExtra("onlyAdd", true)
         }
+        val marketIntent =
+            Intent(ACTION_VIEW, Uri.parse("market://details?id=$NEXTCLOUD_PACKAGE")).apply {
+                addFlags(FLAG_ACTIVITY_NEW_TASK)
+            }
         val isInstalled = packageManager.resolveActivity(intent, 0) != null
+        val canInstall = packageManager.resolveActivity(marketIntent, 0) != null
         val summaryRes = if (isInstalled) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootFetcher.kt
```diff
@@ -156,11 +156,19 @@ internal class StorageRootFetcher(private val context: Context, private val isRe
...
+            }
         val isInstalled = packageManager.resolveActivity(intent, 0) != null
+        val canInstall = packageManager.resolveActivity(marketIntent, 0) != null
         val summaryRes = if (isInstalled) {
             if (isRestore) R.string.storage_fake_nextcloud_summary_installed
             else R.string.storage_fake_nextcloud_summary_unavailable
-        } else R.string.storage_fake_nextcloud_summary
+        } else {
+            if (!canInstall) R.string.storage_fake_nextcloud_summary_unavailable_market
+            else R.string.storage_fake_nextcloud_summary
```
grote, line 170: Nitpick: I'd switch the logic here and don't negate `canInstall`.

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootFetcher.kt
```diff
@@ -171,15 +179,10 @@ internal class StorageRootFetcher(private val context: Context, private val isRe
             availableBytes = null,
             isUsb = false,
             requiresNetwork = true,
-            enabled = !isInstalled || isRestore,
+            enabled = (!isInstalled && canInstall) || (isRestore && (!isInstalled && canInstall)),
```
grote, line 182: Can we maybe simplify this logic here? Would `isInstalled || canInstall` work as well? Just judging from the click listener here.

grote (changes requested):

chirayudesai (approved):
Works with a store installed as well.

Issue #264, opened 2021-06-12
"Not backed up as it wasn't used recently"? That's not for the backup app to decide!


2019-05-10:
Not backing up stuff b/c "it wasn't used recently" is a no-go and deal breaker.
a) what is "recently"? 
b) why is that not communicated before backup
c) why is there no setting to backup those nevertheless?

rugk:
I kind of agree, though, I like to current behaviour that could be used as an _suggestion_ i.e. do not force the user into doing this, …

However, that all said, I just read the FAQ and "recently used" is likely an understatement. See https://github.com/seedvault-app/seedvault/wiki/FAQ#why-do-apps-not-get-backed-up-when-not-recently-used

I.e. the app was likely not used at all or is empty. Only the "force stop" thing is strange. And it looks like a limitation of the Android backup system, not Seedvault.
Also IIRC, the APK is still baked up, which is quite convenient.

---

As such an **alternative proposal**: Rename that option. 
Maybe something like "No backup as app is not actively used." or even "No backup as app is unused." That could make it more clear by omitting the vague "recently".

Derkades:
I see it only for apps I've never opened before

grote:
https://github.com/seedvault-app/seedvault/wiki/FAQ#why-do-apps-not-get-backed-up-when-not-recently-used

grote:
I think the main issue is that the backup policy here is just hard to communicate to users. I wonder how Google Play Services do it as the same policy must apply to their backup service.

gh33291:
This is not a good default, especially because it can't even be configured.
I have many apps that I've disabled, but I care about their app data very much, and none of them are being backed up.

Eridyne:
This is an absolutely comically terrible default.

It makes seedvault silently unreliable, meaning unsafe, for app backup.

PR #308, opened 2021-08-28
Link FAQ in Readme to make it more discoverable

rugk:
E.g. for https://github.com/seedvault-app/seedvault/issues/264 I found the FAQ way to late.

And for users this is a very valuable resource. As such, IMHO you should make it more discoverable, e.g. like this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 
 A backup application for the [Android Open Source Project](https://source.android.com/).
 
+If you are having an issue/question, please look at our [FAQ](../../wiki/FAQ).
+
 ## Components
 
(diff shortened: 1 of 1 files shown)


chirayudesai (approved):

chirayudesai:
Ok for now. In the future I wanted to put everything on the website since the FAQ can only be edited by people with access to the repository (thus isn't not really a wiki anymore, is it? :) )

rugk:
> since the FAQ can only be edited by people with access to the repository (thus isn't not really a wiki anymore, is it? :) )

Well… you can change that and allow editing by anyone…

PR #319, opened 2021-09-29
Allow launching restore through a dialer code

chirayudesai:
* We don't show Restore in menu by default since it's
  not the best idea to restore a running system
* However, at the same time, it's good to have a way to do
  that for those who'd like to restore anyway, and the only
  current way is adb, which is not ideal
* Dialing "*#*#RESTORE#*#*" will launch the restore activity

Change-Id: I258fead82f7e916a4de0b314e1840d7aa4b3746c

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/SecretCodeReceiver.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/SecretCodeReceiver.kt
@@ -1,0 +1,24 @@
+package com.stevesoltys.seedvault
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
+import android.util.Log
+import com.stevesoltys.seedvault.restore.RestoreActivity
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/stevesoltys/seedvault/SecretCodeReceiver.kt
```diff
@@ -0,0 +1,24 @@
...
+class SecretCodeReceiver : BroadcastReceiver() {
+
+	override fun onReceive(context: Context, intent: Intent) {
+		val host = intent.data.host
+		if (!RESTORE_SECRET_CODE.equals(host)) return
+		Log.d(TAG, "Restore secret code received.")
+		val i = Intent(context, RestoreActivity::class.java).apply {
+			flags = FLAG_ACTIVITY_NEW_TASK
+		}
+	    context.startActivity(i)
```
grote, line 22: white-space nitpick. ktlint should catch this and fail

grote (dismissed):

grote (approved):

PR #325, opened 2021-10-01
Fix translations

mikeNG:


chirayudesai (approved):
My bad. Mind picking this over to 12 as well?

PR #327, opened 2021-10-07
Upgrade Format for Android 12

### LINKED ISSUE #127: Backup format

grote:
This uses the same (faster and more secure) crypto that storage backups use. The storage layout was simplified to use random names inside a single folder. Backups made with old 1.x or 2.x Seedvault can still be restored, but new backups will be made with a new format.

Biggest problem is for legacy installs that did not store the new main key yet. Those will need to generate a new recovery code for backups to work.

K/V backups have been changed to use a compressed sqllite database that is uploaded for changes instead of saving many tiny files. This speeds up backups for `@pm@` and call log significantly.

Closes #127


chirayudesai:
> Biggest problem is for legacy installs that did not store the new main key yet.

How legacy are we talking here? Let's say in SeedVault version, 11-1.0? 11-2.0?

> Those will need to generate a new recovery code for backups to work.

Will they be prompted for this or will it have to be manual?

app/build.gradle
```diff
@@ -105,6 +105,8 @@ dependencies {
     implementation rootProject.ext.std_libs.androidx_documentfile
     implementation rootProject.ext.std_libs.com_google_android_material
 
+    implementation rootProject.ext.storage_libs.com_google_crypto_tink_android
```
chirayudesai, line 108: Will need to update `Android.bp`

chirayudesai (commented):
Haven't reviewed code yet.

Will test.

grote:
> How legacy are we talking here? Let's say in SeedVault version, 11-1.0? 11-2.0?

Everything that got the recovery code before https://github.com/seedvault-app/seedvault/commit/851407037e2c9690d6a16272d5691fff79fd4cd1

> >    Those will need to generate a new recovery code for backups to work.

> Will they be prompted for this or will it have to be manual?

They should get a persistent notification making them aware of the issue.

app/build.gradle
grote: This is already included because storage lib is using it already.

chirayudesai (approved):
Works fine from initial testing.

New backups are quick, and restoring an old backup still works as well. The speed up is apparent if you try to take another backup right after that.

Needs more testing, we can do that after merge and before tagging stable.

PR #329, opened 2021-10-08
Android12 bring up

grote:
Draft only because based on #327


contactsbackup/build.gradle
```diff
@@ -43,7 +48,7 @@ android {
 }
 
 def aospDeps = fileTree(include: [
-        // out/target/common/obj/JAVA_LIBRARIES/com.android.vcard_intermediates/classes.jar
+        // out/target/common/obj/JAVA_LIBRARIES/com.android.vcard_intermediates_classes.jar
```
chirayudesai: No no, it's is classes.jar still, I only renamed the file for upload.

gradle/dependencies.gradle
```diff
@@ -17,9 +17,9 @@ ext.aosp_libs = fileTree(include: [
     // framework_intermediates/classes-header.jar works for gradle build as well,
     // but not unit tests, so we use the actual classes (without updatable modules).
     //
-    // out/target/common/obj/JAVA_LIBRARIES/framework-minus-apex_intermediates/classes.jar
+    // out/target/common/obj/JAVA_LIBRARIES/framework-minus-apex_intermediates_classes.jar
     'android.jar',
-    // out/target/common/obj/JAVA_LIBRARIES/core-libart.com.android.art.release_intermediates/classes.jar
+    // out/target/common/obj/JAVA_LIBRARIES/core-libart.com.android.art_intermediates_classes.jar
```
chirayudesai: Same, keep this `/classes.jar` like before.

app/src/main/java/com/stevesoltys/seedvault/crypto/KeyManager.kt
```diff
@@ -100,7 +100,9 @@ internal class KeyManagerImpl(
             .setEncryptionPaddings(ENCRYPTION_PADDING_NONE)
             .setRandomizedEncryptionRequired(true)
         // unlocking is required only for decryption, so when restoring from backup
-        builder.setUnlockedDeviceRequired(true)
+        // FIXME disabled for Android 12 GSI as it crashes when importing the key
```
chirayudesai, line 103: For stuff like this I tend to add [DO NOT MERGE] or something to the commit message so it's easy to spot.

grote: I didn't do this, because it is not the end of the world if this doesn't get re-enabled. This key will only be used to restore legacy backups anyway.

chirayudesai (approved):
Builds in AOSP, boots.

Backup works, so does restore, even with an old backup.
Quick test of upgrade from 11 build worked too.

Needs more testing, can continue doing that before we cut a stable tag.

Testing with GSI for now.

PR #332, opened 2021-10-13
Clean up stuck notifications when service gets killed

grote:
Cancel left-over notifications that are still ongoing.

We have seen a race condition where the service was taken down at the same time as `BackupObserver#backupFinished()` was called, early enough to miss the cancel.

This won't bring back the expected finish notification in this case, but at least we don't leave stuck notifications laying around.


grote:
We have only seen this on Android 12 so far, so maybe no need to backport this to 11.

chirayudesai (approved):

chirayudesai:
I haven't been able to repro, but this looks good.

PR #341, opened 2021-11-12
Allow DAVx5 DocumentsProvider by default

grote:
Works better than the Nextcloud app. This allows more people to test. OEMs as always need to make sure that people can restore with this.


chirayudesai:
Needs DAVx⁵ 4.0 at least - https://forums.bitfire.at/post/14525
* https://play.google.com/store/apps/details?id=at.bitfire.davdroid
* https://f-droid.org/en/packages/at.bitfire.davdroid/
* https://www.davx5.com/download

chirayudesai (approved):

PR #343, opened 2021-11-24
Move Seedvault to /system_ext partition

mikeNG:
Change-Id: I5a1d33dec45ade9a30925b471652878bfa39404f

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -43,6 +43,7 @@ android_app {
     manifest: "app/src/main/AndroidManifest.xml",
 
     platform_apis: true,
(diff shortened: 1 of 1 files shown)


mikeNG:
This makes android build happy on Pixel devices where PRODUCT_ENFORCE_ARTIFACT_PATH_REQUIREMENTS is enabled on 12. Previously we tried /product partition but that didn't have hidden API access, however /system_ext works for that case.

grote (approved):

chirayudesai (approved):

PR #359, opened 2022-01-04
RestoreFiles: Pass on RESULT_SKIP

chirayudesai:
* Grabbed from commit "Seedvault: redesign SUW activities"
  I2effdf5f11c8d92899f328abf0027e7bfd1a57ba

Change-Id: Ie81407ee79c16d6c40a352efa846a05eac1ada37

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
@@ -1,5 +1,6 @@
 package com.stevesoltys.seedvault.restore
 
+import android.app.Activity.RESULT_FIRST_USER
 import android.app.Activity.RESULT_OK
 import android.os.Bundle
 import android.view.LayoutInflater
@@ -34,7 +35,9 @@ internal class RestoreFilesFragment : SnapshotFragment() {
         val skipView: TextView = footer.findViewById(R.id.skipView)
         skipView.setOnClickListener {
             requireActivity().apply {
-                setResult(RESULT_OK)
+                // Equivalent to com.google.android.setupcompat.util.ResultCodes.RESULT_SKIP
+                // SetupWizard handles this
+                setResult(RESULT_FIRST_USER)
                 finishAfterTransition()
             }
         }


app/src/main/java/com/stevesoltys/seedvault/restore/RestoreSetFragment.kt
```diff
@@ -50,6 +51,10 @@ class RestoreSetFragment : Fragment() {
 
         skipView.setOnClickListener {
             viewModel.onFinishClickedAfterRestoringAppData()
+            // Equivalent to com.google.android.setupcompat.util.ResultCodes.RESULT_SKIP
+            // SetupWizard handles this
+            requireActivity().setResult(RESULT_FIRST_USER)
+            requireActivity().finishAfterTransition()
```
grote: If we finish here, how will the user be able to restore files?

chirayudesai: Good point. 

https://github.com/seedvault-app/seedvault/blob/129f8d90e5862a3614f11e187d85ca1022e08701/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt#L35-L38

Could change the `RESULT_OK` there.

And I just noticed that uses `.apply()`, which is idiomatic kotlin style yes?

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
```diff
@@ -34,7 +35,9 @@ internal class RestoreFilesFragment : SnapshotFragment() {
         val skipView: TextView = footer.findViewById(R.id.skipView)
         skipView.setOnClickListener {
             requireActivity().apply {
-                setResult(RESULT_OK)
+                // Equivalent to com.google.android.setupcompat.util.ResultCodes.RESULT_SKIP
+                // SetupWizard handles this
+                setResult(RESULT_FIRST_USER)
                 finishAfterTransition()
```
grote, line 41: I don't really remember why I was setting `RESULT_OK` here before. Looks like Seedvault itself isn't caring about the result, so I can only think of SUW in which case the new result should be better. Or how does SUW treat result OK? Move to the next step as well?

Asking, because there's also `RestoreFilesStartedFragment` below in the same file which sets a result.

chirayudesai: > Or how does SUW treat result OK? Move to the next step as well?

Yes but for that it does a transition I believe.

https://review.lineageos.org/c/LineageOS/android_packages_apps_SetupWizard/+/315577/2/src/org/lineageos/setupwizard/backup/RestoreIntroActivity.java

This handles RESULT_SKIP

RESULT_OK would go to https://github.com/LineageOS/android_packages_apps_SetupWizard/blob/lineage-19.0/src/org/lineageos/setupwizard/SubBaseActivity.java#L129


I only tested this change with skipping so far, not with a full restore.

chirayudesai:
Will be included in https://github.com/seedvault-app/seedvault/pull/392

chirayudesai:
Let's revisit this with https://github.com/seedvault-app/seedvault/pull/402 open now, in case this is still needed.

chirayudesai:
I don't think we need this anymore.

PR #364, opened 2022-01-10
Always add a storage option placeholder for DavX5

grote:
This MR will allow the user to install/setup DavX5 during the phones' SetupWizard when normal operation is not yet allowed. This is important for ROMs that only allow restore at that point.

The MR also refactors the `StorageRoots` in a dedicated commit to be more generic to prepare for non-SAF storage plugins.


grote:
@rfc2822 you might want to have a look at that second commit and note that we rely on `at.bitfire.davdroid.ui.webdav.WebdavMountsActivity` being exported now. You might also want to consider running (part?) of your in-app setup wizard when that activity gets launched directly.

rfc2822:
Hi,

I'll have a look when I have some minutes.

chirayudesai (approved):
Perfect, thanks!

PR #366, opened 2022-01-11
Deleted translation using Weblate (Uyghur)

mikeNG:


chirayudesai (approved):

chirayudesai:
Note: This was deleted since AOSP does not support this language, so there's not much use of translating just Seedvault.

Adding a new language to AOSP is not trivial, since there's thousands of strings to translate.

Issue #77, opened 2020-03-09
Backup work profile (Shelter)


ypid:
Hi,

I have setup a Android work profile so when I open System -> Backup, the settings app asks in which profile the Seedvault should be opened. Personal profile works as intended. But when I select "Work", it says "Disabled by admin". I am the admin and I did not disable it :) I did a quick grep over `frameworks/base`, Internet search and logcat. No obvious setting. Do you see a trick to patch Android so that backing up the work profile works?

Keep up the good work!

stevesoltys:
Hey!

I'm thinking this is what you may be looking for: https://github.com/aosp-mirror/platform_packages_apps_settings/blob/master/src/com/android/settings/backup/BackupSettingsHelper.java#L216

Untested, but looks right to me. Let me know if changing that works for you. If it's not there, I would guess it's somewhere around there.

ypid:
Looks promising. I will patch this function to always return true and enable logging in my next build for the 2020-04-05 security patch. For reference, logcat shows `getIntentForBackupSettings()` of that file is being called when I try to open Seedvault for the work profile:

```Shell
$ adb logcat -v year,uid,color -s BackupSettingsHelper
--------- beginning of system
--------- beginning of crash
--------- beginning of main
2020-03-13 22:15:18.781 1101000  7049  7049 E BackupSettingsHelper: Backup transport has not provided an intent or the component for the intent is not found!
```

I will report back.

stevesoltys:
Will close this for now, if you find a good solution I can add it to the documentation so those interested can consider applying the patch.

Thanks for bringing this to our attention.

ypid:
The patch:

```
project packages/apps/Settings/
diff --git a/packages/apps/Settings/src/com/android/settings/backup/BackupSettingsHelper.java b/packages/apps/Settings/src/com/android/     settings/backup/BackupSettingsHelper.java
index 1d3455b149..52b0d143a0 100644
--- a/packages/apps/Settings/src/com/android/settings/backup/BackupSettingsHelper.java
+++ b/packages/apps/Settings/src/com/android/settings/backup/BackupSettingsHelper.java
@@ -211,15 +211,8 @@ public class BackupSettingsHelper {

     /** Checks if backup service is enabled for this user. */
     public boolean isBackupServiceActive() {
-        boolean backupOkay;
-        try {
-            backupOkay = mBackupManager.isBackupServiceActive(UserHandle.myUserId());
-        } catch (Exception e) {
[...]

PR #381, opened 2022-02-09
Allow secondary user backup to USB

Uldiniad:
By default, Android exposes USB devices only to the main user.
In order to query, read and write to it, the signature permission INTERACT_ACROSS_USERS_FULL (optional) is granted to create Seedvault's context as the system user.

Change-Id: I0b1b4c8c5aeeb226419ff94e15f631ebe1db66df
Issue: calyxos#437
Issue: https://github.com/seedvault-app/seedvault/issues/77

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
@@ -24,14 +24,15 @@ internal class RestoreStorageViewModel(
 
     override fun onLocationSet(uri: Uri) {
         viewModelScope.launch(Dispatchers.IO) {
-            saveStorage(uri)
+            val storage = createStorage(uri)
             val hasBackup = try {
-                storagePlugin.hasBackup(uri)
+                storagePlugin.hasBackup(storage)
             } catch (e: IOException) {
                 Log.e(TAG, "Error reading URI: $uri", e)
                 false
             }
             if (hasBackup) {
+                saveStorage(storage)
                 mLocationChecked.postEvent(LocationResult())
             } else {
                 Log.w(TAG, "Location was rejected: $uri")
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
@@ -24,6 +24,7 @@ internal class RestoreStorageViewModel(
 
     override fun onLocationSet(uri: Uri) {
         viewModelScope.launch(Dispatchers.IO) {
+            saveStorage(uri)
             val hasBackup = try {
                 storagePlugin.hasBackup(uri)
             } catch (e: IOException) {
@@ -31,7 +32,6 @@ internal class RestoreStorageViewModel(
                 false
             }
             if (hasBackup) {
-                saveStorage(uri)
                 mLocationChecked.postEvent(LocationResult())
             } else {
                 Log.w(TAG, "Location was rejected: $uri")
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
@@ -32,7 +32,6 @@ internal class RestoreStorageViewModel(
             }
             if (hasBackup) {
                 saveStorage(uri)
-
                 mLocationChecked.postEvent(LocationResult())
             } else {
                 Log.w(TAG, "Location was rejected: $uri")
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
@@ -18,13 +18,12 @@ import android.provider.DocumentsContract.Root.FLAG_SUPPORTS_CREATE
 import android.provider.DocumentsContract.Root.FLAG_SUPPORTS_IS_CHILD
 import android.util.Log
 import com.stevesoltys.seedvault.R
+import com.stevesoltys.seedvault.getSystemContext
 import com.stevesoltys.seedvault.ui.storage.StorageOption.SafOption
 
 internal object StorageRootResolver {
 
     private val TAG = StorageRootResolver::class.java.simpleName
-
-    private const val usbAuthority = "com.android.externalstorage.documents"
 
     fun getStorageRoots(context: Context, authority: String): List<SafOption> {
         val roots = ArrayList<SafOption>()
@@ -37,12 +36,17 @@ internal object StorageRootResolver {
                     if (root != null) roots.add(root)
                 }
             }
-            if (usbAuthority == authority && UserHandle.myUserId() != UserHandle.USER_SYSTEM) {
-                val c: Context = context.createContextAsUser(UserHandle.SYSTEM, 0)
+            // add special system user roots for USB devices
+            val c = context.getSystemContext {
+                authority == AUTHORITY_STORAGE && UserHandle.myUserId() != UserHandle.USER_SYSTEM
+            }
+            // only proceed if we really got a different [Context], e.g. had permission for it
+            if (context !== c) {
                 c.contentResolver.query(rootsUri, null, null, null, null)?.use { cursor ->
                     while (cursor.moveToNext()) {
-                        // Pass in context since it is used to query package manager for app icons
+                        // Pass in [context] since it is used to query package manager for app icons
                         val root = getStorageRoot(context, authority, cursor)
+                        // only add USB storage from system user, no others
                         if (root != null && root.isUsb) roots.add(root)
                     }
                 }
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
@@ -3,6 +3,7 @@ package com.stevesoltys.seedvault.ui.storage
 import android.content.Context
 import android.database.Cursor
 import android.graphics.drawable.Drawable
+import android.os.UserHandle
 import android.provider.DocumentsContract
 import android.provider.DocumentsContract.Root.COLUMN_AVAILABLE_BYTES
 import android.provider.DocumentsContract.Root.COLUMN_DOCUMENT_ID
@@ -23,6 +24,8 @@ internal object StorageRootResolver {
 
     private val TAG = StorageRootResolver::class.java.simpleName
 
+    private const val usbAuthority = "com.android.externalstorage.documents"
+
     fun getStorageRoots(context: Context, authority: String): List<SafOption> {
         val roots = ArrayList<SafOption>()
         val rootsUri = DocumentsContract.buildRootsUri(authority)
@@ -32,6 +35,16 @@ internal object StorageRootResolver {
                 while (cursor.moveToNext()) {
                     val root = getStorageRoot(context, authority, cursor)
                     if (root != null) roots.add(root)
+                }
+            }
+            if (usbAuthority == authority && UserHandle.myUserId() != UserHandle.USER_SYSTEM) {
+                val c: Context = context.createContextAsUser(UserHandle.SYSTEM, 0)
+                c.contentResolver.query(rootsUri, null, null, null, null)?.use { cursor ->
+                    while (cursor.moveToNext()) {
+                        // Pass in context since it is used to query package manager for app icons
+                        val root = getStorageRoot(context, authority, cursor)
+                        if (root != null && root.isUsb) roots.add(root)
+                    }
                 }
             }
         } catch (e: Exception) {
--- a/storage/lib/src/main/java/org/calyxos/backup/storage/plugin/saf/SafStoragePlugin.kt
+++ b/storage/lib/src/main/java/org/calyxos/backup/storage/plugin/saf/SafStoragePlugin.kt
@@ -35,12 +35,15 @@ private const val TAG = "SafStoragePlugin"
 public abstract class SafStoragePlugin(
     private val appContext: Context,
 ) : StoragePlugin {
-
-    private val cache = SafCache()
-    // In the case of USB storage, if INTERACT_ACROSS_USERS_FULL is granted, this context will match
-    // the system user's application context. Otherwise, matches appContext.
+    /**
+     * Attention: This context could be unexpected. E.g. the system user's application context,
+     * in the case of USB storage, if INTERACT_ACROSS_USERS_FULL permission is granted.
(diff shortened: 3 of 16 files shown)


app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
```diff
@@ -31,8 +32,6 @@ internal class RestoreStorageViewModel(
                 false
             }
             if (hasBackup) {
-                saveStorage(uri)
-
```
grote, line 35: You are saving the storage off the I/O thread here, possibly on the UI thread.

However, I don't understand this change in general. If there's an issue here, we should maybe look into it separately. Or is it only your changes in this PR that are surfacing an issue that otherwise doesn't exist?

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
```diff
@@ -34,6 +35,15 @@ internal object StorageRootResolver {
                     if (root != null) roots.add(root)
                 }
             }
+            if (UserHandle.myUserId() != UserHandle.USER_SYSTEM) {
```
grote: Do we need to do this for *all* authorities or could we also check the current authority and only query with system context for authorities that USB storage is using?

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
```diff
@@ -34,6 +35,15 @@ internal object StorageRootResolver {
...
+                val c: Context = context.createContextAsUser(UserHandle.SYSTEM, 0)
+                c.contentResolver.query(rootsUri, null, null, null, null)?.use { cursor ->
+                    while (cursor.moveToNext()) {
+                        val root = getStorageRoot(context, authority, cursor)
```
grote, line 48: `getStorageRoot` root is called with the current user context, not the system context. Is that intentional? Could that have worked in your testing?

storage/lib/src/main/java/org/calyxos/backup/storage/plugin/saf/SafStoragePlugin.kt
```diff
@@ -30,11 +30,12 @@ private const val TAG = "SafStoragePlugin"
 
 @Suppress("BlockingMethodInNonBlockingContext")
 public abstract class SafStoragePlugin(
-    private val context: Context,
+    private val appContext: Context,
 ) : StoragePlugin {
 
     private val cache = SafCache()
     protected abstract val root: DocumentFile?
+    protected abstract val context: Context
```
grote, line 44: Please add JavaDocs to both contexts. This is a library and it isn't obvious what those contexts are and where there are two of them.

grote (changes requested):
Looks good. I found some minor stuff and will do one last pass before merging, checking all the context usage in the places you touched to make sure we are not accidentally using the wrong context somewhere.

app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
Uldiniad: There were errors cause by the code before it trying to read the value and it not being present in `SettingsManager.getStorage`. I don't have the log currently. Additionally the BackupStorageViewMode (https://github.com/seedvault-app/seedvault/blob/android12/app/src/main/java/com/stevesoltys/seedvault/ui/storage/BackupStorageViewModel.kt#L35)l stores it early as well.
It is very likely tied to this PR and not a general issue

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
Uldiniad: This context is sent to query package manager for icons. You mentioned you'd prefer using the context of the current user for package manager?

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
Uldiniad: I am ok with limiting this to USB authorities only, but I don't know how we can identify them

storage/lib/src/main/java/org/calyxos/backup/storage/plugin/saf/SafStoragePlugin.kt
Uldiniad: Done

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
grote: Have you tried checking the existing log or adding explicit authority logging here looking for a pattern to the authority? I'd imagine it to always be the same one e.g. `com.android.externalstorage.documents`.

grote: If we do this only for a specific authority, maybe this isn't even the best location to do it as this code is called for all authorities. Maybe an earlier place in the calling hierarchy of this method would be more appropriate?

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
grote: Ah ok, it wasn't obvious from the diff that the context is only used to fetch the icon. That's fetching a drawable from the current app in this case, so it probably should use app context, but I'd add a comment explaining why using the app context here and not `c`.

app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
grote: First of all, saving the storage off the I/O thread here, possibly on the UI thread is an error that should get resolved in any case. No disk I/O on the UI thread! (If this is happening in other places of the code, that's an error there as well and should get fixed eventually.)

Then, we can't simply move this call around without fully understanding the implications. Just because the backup flow does it, doesn't mean we can do it here as well. If the logic is the same in both flows, why aren't both view models sharing this in common code in `StorageViewModel` then? Just a programming oversight?

I myself don't remember, but looking just at this method here I wonder, what if there are no backups in the location the user has chosen? Will it always be possible to try a different location even if the first one was already saved? Did you try that?

Maybe we don't even need to move the saving call. If we knew what the issue was here, we might be able to come up with a different solution.

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
grote: Could factor this `if` part out into its own method and call this in the loop in `getStorageOptions()` only if the authority matches the one we care about.

app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
grote: It looks the only issue that you might have run into is in the initialization of `DocumentsProviderStoragePlugin` which tries to access `storage` at construction time now (which might actually cause issues in other places if storage is not yet available). However, if `storage` is still `null` there, then there shouldn't be any crashes or anything. But it might have happened that you couldn't access the system user USB drive from a work profile at restore time.

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
Uldiniad: Would this negatively affect `openDocumentTree` in `StorageActivity` since it also calls `getStorageRoots()` ?

grote: Good point, that's the fall-back for selecting storage without MANAGE_DOCUMENTS permissions. It only works on specific Uris selected by the user though. So how would the work-profile user select system usb storage via the file picker? If that's not possible anyway, I think we can ignore this issue.

Uldiniad: I just tried attaching a USB and adding an attachment from the USB using the file picker to a conversation on the messaging app in the work profile and it worked. Seems like this usecase needs handling

grote: Wow, so the files app has code for this as well? So you can access USB devices from work profile after all?

Uldiniad: Yeah. There are various intents that work across profiles. USB_DEVICE_ATTACHED is one of them. Additionally the files app has some internal code that limits exposure to other profiles.
So we have to handle this within the getStorageRoots function given this scenario?

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
Uldiniad: Done

app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
Uldiniad: Moved it back. Will post log when I have it

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
Uldiniad: Added check for `com.android.externalstorage.documents`

Uldiniad:
```
02-16 16:35:59.761 12887 12990 E DocumentsStorage: Error finding file blocking
02-16 16:35:59.761 12887 12990 E DocumentsStorage: java.io.IOException
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at com.stevesoltys.seedvault.plugins.saf.DocumentsStorageKt$getLoadedCursor$2.invokeSuspend(DocumentsStorage.kt:281)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at com.stevesoltys.seedvault.plugins.saf.DocumentsStorageKt$getLoadedCursor$2.invoke(Unknown Source:8)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at com.stevesoltys.seedvault.plugins.saf.DocumentsStorageKt$getLoadedCursor$2.invoke(Unknown Source:4)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturnIgnoreTimeout(Undispatched.kt:102)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at kotlinx.coroutines.TimeoutKt.setupTimeout(Timeout.kt:78)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at kotlinx.coroutines.TimeoutKt.access$setupTimeout(Timeout.kt:1)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at kotlinx.coroutines.TimeoutKt.withTimeout(Timeout.kt:31)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at com.stevesoltys.seedvault.plugins.saf.DocumentsStorageKt.getLoadedCursor(DocumentsStorage.kt:279)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at com.stevesoltys.seedvault.plugins.saf.DocumentsStorageKt.getLoadedCursor$default(DocumentsStorage.kt:278)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at com.stevesoltys.seedvault.plugins.saf.DocumentsStorageKt.listFilesBlocking(DocumentsStorage.kt:204)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at com.stevesoltys.seedvault.plugins.saf.DocumentsStorageKt.findFileBlocking(DocumentsStorage.kt:248)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at com.stevesoltys.seedvault.plugins.saf.DocumentsProviderStoragePlugin.hasBackup(DocumentsProviderStoragePlugin.kt:82)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at com.stevesoltys.seedvault.ui.storage.RestoreStorageViewModel$onLocationSet$1.invokeSuspend(RestoreStorageViewModel.kt:28)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:241)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:594)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at kotlinx.coroutines.scheduling.CoroutineScheduler.access$runSafely(CoroutineScheduler.kt:60)
02-16 16:35:59.761 12887 12990 E DocumentsStorage:      at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:740)
```
this is the error when save storage is not executed

app/src/main/java/com/stevesoltys/seedvault/ui/storage/RestoreStorageViewModel.kt
chirayudesai: Bump.

@Uldiniad do one thing, split this part into a separate commit - just this one liner.

Uldiniad: Done

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
```diff
@@ -34,6 +36,21 @@ internal object StorageRootResolver {
                     if (root != null) roots.add(root)
                 }
             }
+            // add special system user roots for USB devices
+            val c = context.getSystemContext {
+                authority == AUTHORITY_STORAGE && UserHandle.myUserId() != UserHandle.USER_SYSTEM
+            }
+            // only proceed if we really got a different [Context], e.g. had permission for it
+            if (context !== c) {
```
grote, line 44: I am not a 100% sure that this will work as expected. Mainly did it like this, so we don't run into a situation where the permission isn't granted and it uses the same context creating duplicate roots.

chirayudesai: Seems to work fine.

chirayudesai (approved):

PR #392, opened 2022-03-24
Add SetupWizard libs from AOSP

chirayudesai:
TODO: Add submodule instructions to README or wiki

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
@@ -1,5 +1,6 @@
 package com.stevesoltys.seedvault.restore
 
+import android.app.Activity.RESULT_FIRST_USER
 import android.app.Activity.RESULT_OK
 import android.os.Bundle
 import android.view.LayoutInflater
@@ -34,7 +35,9 @@ internal class RestoreFilesFragment : SnapshotFragment() {
         val skipView: TextView = footer.findViewById(R.id.skipView)
         skipView.setOnClickListener {
             requireActivity().apply {
-                setResult(RESULT_OK)
+                // Equivalent to com.google.android.setupcompat.util.ResultCodes.RESULT_SKIP
+                // SetupWizard handles this
+                setResult(RESULT_FIRST_USER)
                 finishAfterTransition()
             }
         }
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
@@ -1,16 +1,18 @@
 package com.stevesoltys.seedvault.restore
 
 import android.app.Activity.RESULT_OK
+import android.graphics.drawable.Drawable
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.view.ViewStub
 import android.widget.Button
-import android.widget.TextView
 import androidx.fragment.app.Fragment
 import com.google.android.setupcompat.util.ResultCodes.RESULT_SKIP
+import com.google.android.setupdesign.GlifLayout
 import com.stevesoltys.seedvault.R
+import com.stevesoltys.seedvault.ui.getColorAccent
 import org.calyxos.backup.storage.api.SnapshotItem
 import org.calyxos.backup.storage.ui.restore.SnapshotFragment
 import org.koin.androidx.viewmodel.ext.android.sharedViewModel
@@ -27,12 +29,16 @@ internal class RestoreFilesFragment : SnapshotFragment() {
 
         val topStub: ViewStub = v.findViewById(R.id.topStub)
         topStub.layoutResource = R.layout.header_snapshots
-        topStub.inflate()
+        val header = topStub.inflate()
+        val suwLayout: GlifLayout = header.findViewById(R.id.setup_wizard_layout)
+        val icon: Drawable = suwLayout.getIcon()
+        icon.setTintList(getColorAccent(requireContext()))
+        suwLayout.setIcon(icon)
 
         val bottomStub: ViewStub = v.findViewById(R.id.bottomStub)
         bottomStub.layoutResource = R.layout.footer_snapshots
         val footer = bottomStub.inflate()
-        val skipView: TextView = footer.findViewById(R.id.skipView)
+        val skipView: Button = footer.findViewById(R.id.skipView)
         skipView.setOnClickListener {
             requireActivity().apply {
                 // SetupWizard handles this
@@ -55,6 +61,11 @@ internal class RestoreFilesStartedFragment : Fragment() {
         savedInstanceState: Bundle?,
     ): View {
         val v: View = inflater.inflate(R.layout.fragment_restore_files_started, container, false)
+
+        val suwLayout: GlifLayout = v.findViewById(R.id.setup_wizard_layout)
+        val icon: Drawable = suwLayout.getIcon()
+        icon.setTintList(getColorAccent(requireContext()))
+        suwLayout.setIcon(icon)
 
         val button: Button = v.findViewById(R.id.button)
         button.setOnClickListener {
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
@@ -1,6 +1,5 @@
 package com.stevesoltys.seedvault.restore
 
-import android.app.Activity.RESULT_FIRST_USER
 import android.app.Activity.RESULT_OK
 import android.os.Bundle
 import android.view.LayoutInflater
@@ -10,6 +9,7 @@ import android.view.ViewStub
 import android.widget.Button
 import android.widget.TextView
 import androidx.fragment.app.Fragment
+import com.google.android.setupcompat.util.ResultCodes.RESULT_SKIP
 import com.stevesoltys.seedvault.R
 import org.calyxos.backup.storage.api.SnapshotItem
 import org.calyxos.backup.storage.ui.restore.SnapshotFragment
@@ -35,9 +35,8 @@ internal class RestoreFilesFragment : SnapshotFragment() {
         val skipView: TextView = footer.findViewById(R.id.skipView)
         skipView.setOnClickListener {
             requireActivity().apply {
-                // Equivalent to com.google.android.setupcompat.util.ResultCodes.RESULT_SKIP
                 // SetupWizard handles this
-                setResult(RESULT_FIRST_USER)
+                setResult(RESULT_SKIP)
                 finishAfterTransition()
             }
         }
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/UiUtils.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/UiUtils.kt
@@ -1,6 +1,7 @@
 package com.stevesoltys.seedvault.ui
 
 import android.content.Context
+import android.content.res.ColorStateList
 import android.text.format.DateUtils.MINUTE_IN_MILLIS
 import android.text.format.DateUtils.getRelativeTimeSpanString
 import com.stevesoltys.seedvault.R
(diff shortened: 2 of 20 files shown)


chirayudesai:
It doesn't compile due to https://github.com/seedvault-app/setupcompat/blob/13644ae60092b3469c81dad57d58520916f7cb60/build.gradle#L19

Adding it there worked however. Also, setupdesign has it set to 28.

@grote what's the best way to deal with this?

chirayudesai:
Other build issues:

`Caused by: java.lang.RuntimeException: Error while processing seedvault/setupdesign/main/res/drawable-anydpi-v21/sud_navbar_ic_next.xml Width (0) and height (0) cannot be <= 0`

It has width and height set to `@dimen/sud_navbar_ic_intrinsic_size`, which is defined in `setupdesign/main/res/values/dimens.xml`

chirayudesai:
Builds with:

* https://github.com/seedvault-app/setupcompat/pull/1
* https://github.com/seedvault-app/setupdesign/pull/1

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
```diff
@@ -26,15 +29,20 @@ internal class RestoreFilesFragment : SnapshotFragment() {
 
         val topStub: ViewStub = v.findViewById(R.id.topStub)
         topStub.layoutResource = R.layout.header_snapshots
-        topStub.inflate()
+        val header = topStub.inflate()
+        val suwLayout: GlifLayout = header.findViewById(R.id.setup_wizard_layout)
+        val icon: Drawable = suwLayout.getIcon()
```
grote, line 34: If you open this in Android Studio, it tells you to use kotlin syntax instead: `suwLayout.icon`.
 Same below in this file.

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
```diff
@@ -26,15 +29,20 @@ internal class RestoreFilesFragment : SnapshotFragment() {
...
+        icon.setTintList(getColorAccent(requireContext()))
+        suwLayout.setIcon(icon)
```
grote, line 36: If you open this in Android Studio, it tells you to use kotlin syntax instead: `suwLayout.icon = icon`.
Same below in this file.

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreFilesFragment.kt
```diff
@@ -54,6 +62,11 @@ internal class RestoreFilesStartedFragment : Fragment() {
     ): View {
         val v: View = inflater.inflate(R.layout.fragment_restore_files_started, container, false)
 
+        val suwLayout: GlifLayout = v.findViewById(R.id.setup_wizard_layout)
+        val icon: Drawable = suwLayout.getIcon()
+        icon.setTintList(getColorAccent(requireContext()))
```
grote, line 67: Can't the `GlifLayout` handle icon tinting in a better way? Do we have to do these acrobatics here just to apply a tint? As this code change happens a lot, maybe worth looking into this? There's a `sudIconTint` attribute for example and a special theme. Also a `IconMixin` with a `setIconTint()` method which might help.

app/src/main/res/layout/fragment_restore_progress.xml
```diff
@@ -1,84 +1,65 @@
...
-        app:layout_constraintTop_toBottomOf="@+id/progressBar"
-        app:srcCompat="@drawable/ic_cloud_download"
-        app:tint="?android:colorAccent"
-        tools:ignore="ContentDescription" />
+        <ProgressBar
+            android:id="@+id/progressBar"
+            style="?android:attr/progressBarStyleHorizontal"
+            android:layout_width="0dp"
+            android:layout_height="4dp"
+            android:layout_marginTop="16dp"
```
grote, line 22: The progress was neatly anchored at the top before, now there's this empty space above. Is this intentional?

app/src/main/java/com/stevesoltys/seedvault/ui/UiUtils.kt
```diff
@@ -13,3 +14,20 @@ fun Long.toRelativeTime(context: Context): CharSequence {
         getRelativeTimeSpanString(this, now, MINUTE_IN_MILLIS, 0)
     }
 }
+
+// Ported from
+// https://cs.android.com/android/platform/superproject/+/android-12.0.0_r1:frameworks/base/packages/SettingsLib/src/com/android/settingslib/Utils.java
+fun getColorAccent(context: Context): ColorStateList? {
```
grote, line 20: Is this maybe available in the libs we are including as submodule now?

app/src/main/java/com/stevesoltys/seedvault/ui/UiUtils.kt
```diff
@@ -13,3 +14,20 @@ fun Long.toRelativeTime(context: Context): CharSequence {
...
+    return getColorAttr(context, android.R.attr.colorAccent)
+}
+
+fun getColorAttr(context: Context, attr: Int): ColorStateList? {
+    val typedArray = context.obtainStyledAttributes(IntArray(1) { attr })
+    var stateList: ColorStateList? = null
```
grote, line 26: `= null` is redundant and can be removed. Could the `var` also be a `val`?

app/src/main/res/layout/fragment_restore_files_started.xml
```diff
@@ -1,56 +1,42 @@
...
-        app:layout_constraintTop_toBottomOf="@+id/imageView" />
+        <TextView
+            android:id="@+id/infoView"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_margin="16dp"
+            android:text="@string/restore_storage_in_progress_info"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/titleView"
```
grote, line 25: There is no more `titleView` in this layout. You would need to use something like `app:layout_constraintTop_toTopOf="parent"` now.

app/src/main/res/layout/fragment_restore_files_started.xml
```diff
@@ -1,56 +1,42 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<com.google.android.setupdesign.GlifLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
```
grote, line 5: Looks like this isn't used anymore in the XMl file.

app/src/main/res/layout/fragment_restore_progress.xml
```diff
@@ -1,84 +1,65 @@
...
-<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<com.google.android.setupdesign.GlifLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/setup_wizard_layout"
     android:layout_width="match_parent"
-    android:layout_height="match_parent">
+    android:layout_height="match_parent"
+    android:icon="@drawable/ic_cloud_download"
```
grote, line 9: Unfortunately, the layout preview in Android Studio isn't working for that `GlifLayout`, so I can't see how this looks in practice.

app/src/main/res/layout/fragment_restore_set.xml
```diff
@@ -1,84 +1,73 @@
...
-        app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@+id/imageView" />
+        <TextView
+            android:id="@+id/titleView"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:text="@string/restore_choose_restore_set"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/imageView"
```
grote, line 24: `imageView` doesn't exist anymore in this layout, the constraint would need to be adapted.

app/src/main/res/layout/fragment_restore_set.xml
```diff
@@ -1,84 +1,73 @@
...
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        style="@style/SudContentFrame">
 
-    <TextView
-        android:id="@+id/titleView"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_margin="16dp"
```
grote, line 24: The margin is gone in the new version. Is this intentional?

app/src/main/res/layout/fragment_restore_set.xml
```diff
@@ -1,84 +1,73 @@
...
 
-    <androidx.recyclerview.widget.RecyclerView
-        android:id="@+id/listView"
-        android:layout_width="match_parent"
-        android:layout_height="0dp"
-        android:layout_marginStart="16dp"
-        android:layout_marginTop="16dp"
-        android:layout_marginEnd="16dp"
-        app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
-        app:layout_constraintBottom_toTopOf="@+id/skipView"
```
grote, line 39: This bottom constraint got removed. Intentionally? I am skeptical that `android:layout_height="0dp"` will work like this.

Also, now the list is most likely overlaying (or underlaying) the skip view.

app/src/main/res/layout/fragment_restore_set.xml
```diff
@@ -1,84 +1,73 @@
...
 
-    <TextView
-        android:id="@+id/errorView"
-        android:layout_width="0dp"
-        android:layout_height="wrap_content"
-        android:layout_margin="16dp"
-        android:textColor="?android:colorError"
-        android:textSize="18sp"
-        android:visibility="invisible"
-        app:layout_constraintBottom_toTopOf="@+id/skipView"
```
grote, line 63: This is missing in the new layout most likely causing the view to jump to the top instead of being vertically centered.

app/src/main/res/layout/fragment_restore_set.xml
```diff
@@ -1,84 +1,73 @@
...
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/titleView"
+            tools:listitem="@layout/list_item_restore_set" />
 
-    <ProgressBar
-        android:id="@+id/progressBar"
-        style="?android:progressBarStyleLarge"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        app:layout_constraintBottom_toTopOf="@+id/skipView"
```
grote, line 50: This is missing in the new layout most likely causing the view to jump to the top instead of being vertically centered.

grote (changes requested):
I stopped half way through reviewing this, because I kept coming across the same issues. Also, since layout previews don't work, could I please get before and after screenshots?

To be honest, I am also a bit skeptical pulling in these two libraries as submodules is worth it for the relatively minor UI changes they bring while at the same time losing tooling support for the parts they do bring.

app/src/main/java/com/stevesoltys/seedvault/ui/UiUtils.kt
chirayudesai: No, SettingsLib is it's own behemoth and not something that's easy to include (nor should we)

chirayudesai:
> I stopped half way through reviewing this, because I kept coming across the same issues. Also, since layout previews don't work, could I please get before and after screenshots?
> 

Will do. For testing I've been `adb installing` this new APK, and then `*#*#RESTORE#*#*`

> To be honest, I am also a bit skeptical pulling in these two libraries as submodules is worth it for the relatively minor UI changes they bring while at the same time losing tooling support for the parts they do bring.

We can re-consider your previous idea of just doing the same layout ourselves directly without these libs, and we can always copy paste certain stuff as needed (dimens / colors / etc)

chirayudesai:
Before:
Screenshot_20220407-033734

After:
Screenshot_20220407-020737_Seedvault


The after is how the rest of the SetupWizard will look going forward (it's close to what Google's looks like), and thus Seedvault restore UI should match that.

chirayudesai:
Closing this, we should implement the UI directly without the setuplibs to

1. Keep the layouts functional in Android Studio
2. Avoid additional dependencies.

It just has to look the same to make it seamless with the SetupWizard, doesn't need to use the same libs.

The lib and Seedvault are both Apache2 so we can freely copy paste things as needed.

PR #402, opened 2022-05-03
Seedvault: Redesign SUW activities for 12

BadDaemon:
* New SetupWizard uses external/setupdesign (sud) and
  external/setupcompat (suc)
  to achieve the desired looks
* It prevents proper previews when included in any app, though, so
  we better try achieving the same by manually styling the activities
  for 12
* This patch mimics the design by applying styles similar to the ones
  from sud, sometimes reusing parts taken from there

Change-Id: Iccf588e5239aeb00645f97b095910986f32c52d0


chirayudesai:
Screenshots:

<details>
<summary>Light</summary>

image
image
image
image
image
</details>

<details>
<summary>Dark / Black:</summary>

image
image
image
image
image
image
image

</details>

grote:
Thanks a lot @BadDaemon that's great stuff! I notice that the file/backup lists now look a bit squeezed with big margin. Is this intentional?

Can you give me a week to look into the diff details?

BadDaemon:
I fixed a few issues:
a) Top margin was wrong, it now matches the actual SetupDesign
b) The items in the lists were indeed squeezed too much since I applied the default margin AND the items had padding.
    I removed the padding for these cases
c) I couldn't get rid of the scrollbars. Moving them in with the margin made it look weird. Therefore I applied the right margin as a padding to the items instead so the scrollbars stay on the right
d) Fixed statusbar background color to match the window's

Screenshots:
<details>
<summary>Light</summary>

Light1
Light2
Light3
Light4

</details>

<details>
<summary>Dark/Black</summary>

Dark1
Dark2
Dark3
Dark4

</details>

BadDaemon:
e) Now also fixed the navigation bar color

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreProgressFragment.kt
```diff
@@ -54,7 +54,7 @@ class RestoreProgressFragment : Fragment() {
         appList.apply {
             layoutManager = this@RestoreProgressFragment.layoutManager
             adapter = this@RestoreProgressFragment.adapter
-            addItemDecoration(DividerItemDecoration(context, VERTICAL))
+            addItemDecoration(DividerItemDecoration(context, SHOW_DIVIDER_NONE))
```
grote: Could we also just not add any item decoration or would this look different?

app/src/main/res/layout/footer_snapshots.xml
```diff
@@ -1,17 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.constraintlayout.widget.ConstraintLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
```
grote: Nitpick: I don't know how you edit those files, but it would be great if you could use the Android Studio XML auto-format, so in later MRs, there's no unrelated changed introduced in the diffs. I comment here, where it is obvious, but also the sort order below is changed by Android Studio.

grote:
What about the fragment where the user needs to enter their recovery code? Was this left intentionally in the old design?

grote:
What about the files backup items where the user needs to select which backup to restore? Was this left intentionally in the old design?

app/src/main/res/layout/fragment_restore_progress.xml
```diff
@@ -68,17 +68,14 @@
...
-        android:layout_marginEnd="16dp"
-        android:layout_marginBottom="16dp"
+        android:layout_marginEnd="40dp"
         android:enabled="false"
         android:text="@string/restore_next"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toEndOf="parent"
-        app:layout_constraintHorizontal_bias="1.0"
-        app:layout_constraintStart_toStartOf="parent" />
+        app:layout_constraintTop_toBottomOf="@id/appList" />
```
grote: `+` missing?

app/src/main/res/layout/fragment_restore_set.xml
```diff
@@ -19,22 +18,20 @@
...
+        style="@style/SudHeaderTitle"
+        android:layout_width="0dp"
         android:layout_height="wrap_content"
-        android:layout_margin="16dp"
         android:text="@string/restore_choose_restore_set"
-        android:textSize="24sp"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@+id/imageView" />
+        app:layout_constraintTop_toBottomOf="@id/imageView" />
```
grote: Why was the `+` removed?

app/src/main/res/layout/fragment_restore_set.xml
```diff
@@ -19,22 +18,20 @@
...
 
     <androidx.recyclerview.widget.RecyclerView
         android:id="@+id/listView"
+        style="@style/SudContent"
         android:layout_width="match_parent"
         android:layout_height="0dp"
-        android:layout_marginStart="16dp"
-        android:layout_marginTop="16dp"
-        android:layout_marginEnd="16dp"
+        android:layout_marginEnd="0dp"
```
grote, line 33: It feels a bit strange use the new `SudContent` style which adds margin start and end, but then set only the end to 0 here and in some other places. why is this?

app/src/main/res/layout/list_item_storage_root.xml
```diff
@@ -41,7 +41,7 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_marginTop="8dp"
-        android:layout_marginEnd="16dp"
+        android:layout_marginEnd="40dp"
```
grote, line 43: I think this is the answer why you set the end margin of the recyclerview to 0, because the padding it is the items. To be honest, I think it is strange that this is mixed like this. The end padding comes from the items, but the start from the list? Is this an SUW thing?

app/src/main/res/layout/list_item_restore_set.xml
```diff
@@ -11,7 +12,6 @@
         android:id="@+id/imageView"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:layout_marginStart="16dp"
```
grote, line 14: removing this margin here causes the ripple affect to have a sharp border with the icon.

grote (changes requested):
Looks good overall. Just some few nitpicks above. It would be great if the padding/margin on lists could be applied a bit more consistently.

app/src/main/res/layout/fragment_restore_progress.xml
BadDaemon: the '+' in these places isn't correct. it's supposed to be used to create the id, which is done on the id tag of the actual element. references should use it without the '+'. not essential to the change, though

app/src/main/res/layout/fragment_restore_set.xml
BadDaemon: see above

app/src/main/res/layout/fragment_restore_set.xml
BadDaemon: this is due to

c) I couldn't get rid of the scrollbars. Moving them in with the margin made it look weird. Therefore I applied the right margin as a padding to the items instead so the scrollbars stay on the right

app/src/main/res/layout/list_item_storage_root.xml
BadDaemon: same as above, c)

app/src/main/res/layout/list_item_restore_set.xml
BadDaemon: will look into this

app/src/main/res/layout/footer_snapshots.xml
BadDaemon: will look into this. i did it mostly because I don't like the different styles where some elements have their start tag standalone and thngs like this have the scheme in the same line (but a second scheme standalone again). aesthetics. not essential or relevant to the change, ofc

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreProgressFragment.kt
BadDaemon: will see

BadDaemon:
> What about the files backup items where the user needs to select which backup to restore? Was this left intentionally in the old design?

i only changed those i was able to actually see and verify. need to check how i can get those two to show and will adjust them accordingly

storage/lib/src/main/res/layout/item_snapshot.xml
```diff
@@ -41,11 +42,10 @@
             android:id="@+id/sizeView"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:layout_marginTop="8dp"
             app:layout_constraintBottom_toBottomOf="parent"
-            app:layout_constraintEnd_toEndOf="@+id/nameView"
+            app:layout_constraintEnd_toEndOf="parent"
             app:layout_constraintStart_toEndOf="@+id/timeView"
-            app:layout_constraintTop_toBottomOf="@+id/nameView"
+            app:layout_constraintTop_toTopOf="parent"
```
grote, line 48: I wonder if with those changes, a long name could go over the size?

Edit: Ah no, the above takes care of this.

grote (approved):

grote:
Thanks for the patience! Changes look good.

Issue #165, opened 2020-11-18
Option to back up all apps by pretending to be a device to device transport


zroug:
Seedvault is bound to Android's backup API which allows apps to exclude themselves from getting backed up. We want to implement another backup mechanism that will allow us to backup more apps by using an API loophole: pretending to do a device-to-device backup.

* [ ] Handle a larger number of apps getting backed up in a single backup pass
* [ ] Warn about insufficient storage space on backup medium (more likely when backing up more apps without storage quota)
* [ ] Pretend (to the system) to do a device-to-device backup run instead of cloud backups

---

Android 11 ignores the `allowBackup` manifest attribute for device to device transfers (https://developer.android.com/about/versions/11/behavior-changes-11#device-to-device-file-transfer). That means the local backup transport supports it in principle and the apps must support it in principle too. I think it would be a nice addition to Seedvault to make this behavior a (configurable) option. If this is indeed possible of course. This is partly related to #36.

grote:
Seedvault would have to pretend to be a device-to-device backup agent. Not sure that's something we would want to do, but it should be as easy as changing a single line of code.

IzzySoft:
Well, maybe you could consider at least a locally attached drive (Flashdrive, SDCard) as "device" – so it would be a device2device transfer? :smile: A power-user-option behind a bunch of warnings would also be acceptible as long as the goal is reached. Backups should be as complete as possible. Of course this would be welcome to < 11 as well :wink:

chirayudesai:
One thing to note here is that just ignoring this flag won't let all apps actually backup properly.

There's multiple reasons for an app to disallow backups. Sometimes it's about not wanting to send backups to Google, sometimes it's about not wanting unencrypted copies (both Google and Seedvault encrypt, and there's an explicit option for this as well), or maybe something else.

Other times it may be because the app might not actually be able to get backed up this way.

The apps would get backed up when the flag is ignored, and it would even say "backup successful". But then when you actually try to restore the app, it may or may not work. And that'd be worse than not having it backed up at all - because you may have made alternate arrangements knowing that, but now if it looks like backup worked maybe not, only to find out later.

Point being, this would need testing. Testing a device-to-device migration would help, since it ignores this flag - seeing what it backs up / restores, what it skips, and if everything works / something doesn't work would be helpful.

Google's change is only for apps' targeting Android 11, of which there may not be many right now (especially with `allowBackup=false`)

[...]

Issue #374, opened 2022-01-23
Restore via nextcloud fails when seedvault tries to restore nextcloud app


jejb:
This is on Lineageos 17.1 without any gapps (so restore has to follow the procedure outlined in #251).  I get full restore of apps then seedvault begins restoring settings, but when it gets to restoring the settings of nextcloud, the nextcloud app crashes and can't be made to work again without a reboot.  Obviously this halts the restore process.  Logcat just shows a null pointer dereference in nextcloud, but I assume some change in its settings file caused this.

It looks like if seedvault wants to use an external app for backup and restore, it can't restore that app until everything else is finished.  This could also be fixed by implementing a separate webdav backend to seedvault so it could communicate with nextcloud without using an external app as suggested in #317.

grote:
Seedvault should not restore Nextcloud when it is used for restoring as this kills the app. We have code to handle the backup, but didn't foresee people making a backup without Nextcloud and then restoring with Nextcloud.

grote:
As a workaround, maybe you see which file it uses to store nextcloud data in logcat and delete that from backup.

jejb:
On Mon, 2022-01-24 at 10:21 -0800, Torsten Grote wrote:
> Seedvault should not restore Nextcloud when it is used for restoring
> as this kills the app.

Right, that's the problem.

>  We have code to handle the backup, but didn't foresee people making
> a backup without Nextcloud and then restoring with Nextcloud.

To be clear: this backup was made with nextcloud and then restored with
nextcloud.  The only problem is because this is a non gapp phone the
restore can't be done from the first init wizard, it has to be done
from the dialer after nextcloud has been installed by f-droid.

James

grote:
> this backup was made with nextcloud and then restored with nextcloud. 

Then there might be two bugs here. If you do another backup, can you confirm that nextcloud gets backed up while using nextcloud for backup (shouldn't actually work due to same app restart) and that there is only one nextcloud version installed?

jejb:
On Wed, 2022-01-26 at 10:17 -0800, Torsten Grote wrote:
> Then there might be two bugs here. If you do another backup, can you
> confirm that nextcloud gets backed up while using nextcloud for
> backup (shouldn't actually work due to same app restart) and that
> there is only one nextcloud version installed?

Yes, I can confirm both.  (I've had the problem before with the
Nextcloud dev app vs the Nextcloud app so I'm careful to make sure
there's only one nextcloud file handler installed).  In the backup I
[...]

PR #407, opened 2022-05-18
Pretend to be a device to device transfer - fake-d2d

Uldiniad:
https://github.com/seedvault-app/seedvault/issues/165

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -67,6 +67,12 @@
         android:name="android.permission.INTERACT_ACROSS_USERS_FULL"
         tools:ignore="ProtectedPermissions" />
 
+    <!-- This is needed to read if the IGNORE_ALLOW_BACKUP_IN_D2D -->
+    <uses-permission android:name="android.permission.LOG_COMPAT_CHANGE" />
+
+    <!-- This is needed to read if the  -->
+    <uses-permission android:name="android.permission.READ_COMPAT_CHANGE_CONFIG" />
+
     <application
         android:name=".App"
         android:allowBackup="false"
--- a/app/src/main/java/com/stevesoltys/seedvault/metadata/Metadata.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/metadata/Metadata.kt
@@ -17,7 +17,7 @@ data class BackupMetadata(
     internal var time: Long = 0L,
     internal val androidVersion: Int = Build.VERSION.SDK_INT,
     internal val androidIncremental: String = Build.VERSION.INCREMENTAL,
-    internal val deviceName: String = "${Build.MANUFACTURER} ${Build.MODEL}",
+    internal val deviceName: String = "D2D",
     internal val packageMetadataMap: PackageMetadataMap = PackageMetadataMap(),
 )
 
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
@@ -15,12 +15,16 @@ import android.content.Context
 import android.content.pm.PackageInfo
 import android.content.pm.PackageManager.GET_SIGNING_CERTIFICATES
 import android.os.ParcelFileDescriptor
+import android.os.UserHandle
+import android.os.UserManager
 import android.util.Log
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.VisibleForTesting.PRIVATE
 import androidx.annotation.WorkerThread
 import com.stevesoltys.seedvault.Clock
 import com.stevesoltys.seedvault.MAGIC_PACKAGE_MANAGER
+import com.stevesoltys.seedvault.PLATFORM_PACKAGE_NAME
+import com.stevesoltys.seedvault.SHARED_BACKUP_AGENT_PACKAGE
 import com.stevesoltys.seedvault.metadata.BackupType
 import com.stevesoltys.seedvault.metadata.MetadataManager
 import com.stevesoltys.seedvault.metadata.PackageState
@@ -143,12 +147,26 @@ internal class BackupCoordinator(
         @Suppress("UNUSED_PARAMETER") isFullBackup: Boolean,
     ): Boolean {
         val packageName = targetPackage.packageName
+        val applicationInfo = targetPackage.applicationInfo
+        // This must be kept in sync with frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
+        if (!targetPackage.allowsBackup(context, UserHandle.myUserId())
+            || (UserHandle.isCore(applicationInfo.uid) &&
+                (applicationInfo.backupAgentName == null ||
+                    (!UserManager.get(context).isSystemUser &&
+                        (packageName == MAGIC_PACKAGE_MANAGER
+                            || packageName == PLATFORM_PACKAGE_NAME))))
+            || packageName == SHARED_BACKUP_AGENT_PACKAGE
+            || applicationInfo.isInstantApp
+            || targetPackage.isStopped()
+            || !applicationInfo.enabled) {
+            return false
+        }
         // Check that the app is not blacklisted by the user
         val enabled = settingsManager.isBackupEnabled(packageName)
         if (!enabled) Log.w(TAG, "Backup of $packageName disabled by user.")
         // We need to exclude the DocumentsProvider used to store backup data.
         // Otherwise, it gets killed when we back it up, terminating our backup.
-        return enabled && targetPackage.packageName != plugin.providerPackageName
+        return enabled && packageName != plugin.providerPackageName
     }
 
     /**
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
@@ -1,6 +1,7 @@
 package com.stevesoltys.seedvault.transport.backup
 
-import android.app.backup.IBackupManager
+import android.Manifest
+import android.app.compat.CompatChanges
 import android.content.Context
 import android.content.pm.ApplicationInfo.FLAG_ALLOW_BACKUP
 import android.content.pm.ApplicationInfo.FLAG_STOPPED
@@ -22,13 +23,15 @@ private val TAG = PackageService::class.java.simpleName
 
 private const val LOG_MAX_PACKAGES = 100
 
+// This must be kept in sync with frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
+private const val IGNORE_ALLOW_BACKUP_IN_D2D = 183147249L
+
 /**
  * @author Steve Soltys
  * @author Torsten Grote
  */
 internal class PackageService(
     private val context: Context,
-    private val backupManager: IBackupManager,
 ) {
 
     private val packageManager: PackageManager = context.packageManager
@@ -50,17 +53,14 @@ internal class PackageService(
                 }
             }
 
-            val eligibleApps =
-                backupManager.filterAppsEligibleForBackupForUser(myUserId, packages.toTypedArray())
-
-            // log eligible packages
-            if (Log.isLoggable(TAG, INFO)) {
-                Log.i(TAG, "Filtering left ${eligibleApps.size} eligible packages:")
-                logPackages(eligibleApps.toList())
-            }
+//            val eligiblePackages = packages.filter { packageInfo ->
+//                isAppEligibleForBackup(packageInfo,
+//                    if (packageInfo.applicationInfo.backupAgentName != null)
+//                        pkg.applicationInfo.flags and FLAG_FULL_BACKUP_ONLY != 0 else true)
+//            }
 
             // add magic @pm@ package (PACKAGE_MANAGER_SENTINEL) which holds package manager data
-            val packageArray = eligibleApps.toMutableList()
+            val packageArray = packages.toMutableList()
             packageArray.add(MAGIC_PACKAGE_MANAGER)
 
             return packageArray.toTypedArray()
@@ -73,7 +73,7 @@ internal class PackageService(
             // because the package info is used by [ApkBackup] which needs signing info.
             return packageManager.getInstalledPackages(GET_SIGNING_CERTIFICATES)
                 .filter { packageInfo ->
-                    packageInfo.doesNotGetBackedUp() && // only apps that do not allow backup
+                    packageInfo.doesNotGetBackedUp(context, myUserId) && // only apps that do not allow backup
                         !packageInfo.isNotUpdatedSystemApp() && // and are not vanilla system apps
                         packageInfo.packageName != context.packageName // not this app
                 }.sortedBy { packageInfo ->
@@ -94,7 +94,7 @@ internal class PackageService(
     val userApps: List<PackageInfo>
         @WorkerThread
         get() = packageManager.getInstalledPackages(GET_INSTRUMENTATION).filter { packageInfo ->
-            packageInfo.isUserVisible(context) && packageInfo.allowsBackup()
+            packageInfo.isUserVisible(context) && packageInfo.allowsBackup(context, myUserId)
         }
 
     /**
@@ -103,7 +103,7 @@ internal class PackageService(
     val userNotAllowedApps: List<PackageInfo>
         @WorkerThread
         get() = packageManager.getInstalledPackages(0).filter { packageInfo ->
-            !packageInfo.allowsBackup() && !packageInfo.isSystemApp()
+            !packageInfo.allowsBackup(context, myUserId) && !packageInfo.isSystemApp()
         }
 
     val expectedAppTotals: ExpectedAppTotals
@@ -114,7 +114,7 @@ internal class PackageService(
             packageManager.getInstalledPackages(GET_INSTRUMENTATION).forEach { packageInfo ->
                 if (packageInfo.isUserVisible(context)) {
                     appsTotal++
-                    if (packageInfo.doesNotGetBackedUp()) {
+                    if (packageInfo.doesNotGetBackedUp(context, myUserId)) {
                         appsOptOut++
                     }
                 }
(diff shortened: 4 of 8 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -65,6 +65,12 @@
         android:name="android.permission.INTERACT_ACROSS_USERS_FULL"
         tools:ignore="ProtectedPermissions" />
 
+    <!-- This is needed to read if the IGNORE_ALLOW_BACKUP_IN_D2D -->
```
Uldiniad, line 70: Better comment needed. Wasn't sure what to write. It is needed to call `CompatChanges.isChangeEnabled(IGNORE_ALLOW_BACKUP_IN_D2D, packageName, UserHandle.of(userId)`

app/src/main/AndroidManifest.xml
```diff
@@ -65,6 +65,12 @@
...
+    <uses-permission android:name="android.permission.LOG_COMPAT_CHANGE" />
+
+    <!-- This is needed to read if the  -->
```
Uldiniad, line 73: Ditto

app/src/main/java/com/stevesoltys/seedvault/metadata/Metadata.kt
```diff
@@ -17,7 +17,7 @@ data class BackupMetadata(
     internal var time: Long = 0L,
     internal val androidVersion: Int = Build.VERSION.SDK_INT,
     internal val androidIncremental: String = Build.VERSION.INCREMENTAL,
-    internal val deviceName: String = "${Build.MANUFACTURER} ${Build.MODEL}",
+    internal val deviceName: String = "D2D",
```
Uldiniad, line 20: This causes the UI to display the backups names as D2D. We should make the UI portion use `${Build.MANUFACTURER} ${Build.MODEL}`

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -50,17 +53,14 @@ internal class PackageService(
...
-            // log eligible packages
-            if (Log.isLoggable(TAG, INFO)) {
-                Log.i(TAG, "Filtering left ${eligibleApps.size} eligible packages:")
-                logPackages(eligibleApps.toList())
-            }
+//            val eligiblePackages = packages.filter { packageInfo ->
+//                isAppEligibleForBackup(packageInfo,
+//                    if (packageInfo.applicationInfo.backupAgentName != null)
+//                        pkg.applicationInfo.flags and FLAG_FULL_BACKUP_ONLY != 0 else true)
+//            }
```
Uldiniad, line 60: Not sure if necessary anymore given the filtering applied in BackupCoordinator's `isAppEligibleForBackup`

app/src/main/AndroidManifest.xml
chirayudesai: Perhaps:

"This is needed to read check if apps are eligible for d2d"

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
```diff
@@ -143,12 +147,26 @@ internal class BackupCoordinator(
         @Suppress("UNUSED_PARAMETER") isFullBackup: Boolean,
     ): Boolean {
         val packageName = targetPackage.packageName
+        val applicationInfo = targetPackage.applicationInfo
+        // This must be kept in sync with frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
```
chirayudesai, line 151: It could be moved to a method.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -173,10 +184,21 @@ internal fun PackageInfo.isNotUpdatedSystemApp(): Boolean {
     return isSystemApp && !isUpdatedSystemApp
 }
 
-internal fun PackageInfo.doesNotGetBackedUp(): Boolean {
+internal fun PackageInfo.doesNotGetBackedUp(context: Context, userId: Int): Boolean {
     if (packageName == MAGIC_PACKAGE_MANAGER || applicationInfo == null) return true
-    return applicationInfo.flags and FLAG_ALLOW_BACKUP == 0 || // does not allow backup
-        applicationInfo.flags and FLAG_STOPPED != 0 // is stopped
+
+    // This must be kept in sync with frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
```
chirayudesai, line 190: This looks the same as the one above so it could be moved to a second method.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -22,13 +23,15 @@ private val TAG = PackageService::class.java.simpleName
 
 private const val LOG_MAX_PACKAGES = 100
 
+// This must be kept in sync with frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
```
chirayudesai, line 26: Maybe use cs.android.com links (pointing to a tag preferably) so we can easily click

That's what we do in some other places.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -157,9 +157,20 @@ internal fun PackageInfo.isSystemApp(): Boolean {
     return applicationInfo.flags and FLAG_SYSTEM != 0
 }
 
-internal fun PackageInfo.allowsBackup(): Boolean {
+internal fun PackageInfo.allowsBackup(context: Context, userId: Int): Boolean {
     if (packageName == MAGIC_PACKAGE_MANAGER || applicationInfo == null) return false
-    return applicationInfo.flags and FLAG_ALLOW_BACKUP != 0
+
+    // This must be kept in sync with frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
+    val isSystemApp = applicationInfo.flags and FLAG_SYSTEM != 0
```
chirayudesai, line 164: Separate discussion but do we want to change this too?

I believe it may need an OS change as well.

I can't think of many system apps this would be useful for, except perhaps Chromium (if it even works)

Otherwise we can easily just directly change the system app, so I think probably best to keep things like they were for system apps (aka what this commit currently has)

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -157,9 +157,20 @@ internal fun PackageInfo.isSystemApp(): Boolean {
...
+    val ignoreAllowBackup = !isSystemApp
+        && (context.checkSelfPermission(Manifest.permission.LOG_COMPAT_CHANGE)
```
chirayudesai, line 166: Is this needed?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -157,9 +157,20 @@ internal fun PackageInfo.isSystemApp(): Boolean {
...
+        == PackageManager.PERMISSION_GRANTED)
+        && (context.checkSelfPermission(Manifest.permission.READ_COMPAT_CHANGE_CONFIG)
```
chirayudesai, line 168: Same as above

chirayudesai (changes requested):
Looks good for the most part, minimal change.

Left some comments, with some tiny nits.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
Uldiniad: The permission check is necessary since it is privileged|signature. Not all deployments would have it granted and calling `CompatChanges.isChangeEnabled(IGNORE_ALLOW_BACKUP_IN_D2D, packageName, UserHandle.of(userId)` without them results in a SecurityException. So we don't ingore allowBackup in those cases

app/src/main/java/com/stevesoltys/seedvault/metadata/Metadata.kt
grote: I guess there's something in AOSP that requires this to be D2D exactly? If so, is there a constant for it?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
grote: Why do we need to re-add these rules from AOSP anyway? Would D2D otherwise backup stuff that for some reason we don't want in our backups?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -22,13 +23,15 @@ private val TAG = PackageService::class.java.simpleName
 
 private const val LOG_MAX_PACKAGES = 100
 
+// This must be kept in sync with frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
+private const val IGNORE_ALLOW_BACKUP_IN_D2D = 183147249L
```
grote, line 27: Where does that come from, can we import it?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -157,9 +157,20 @@ internal fun PackageInfo.isSystemApp(): Boolean {
...
+
+    // This must be kept in sync with frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java
+    val isSystemApp = applicationInfo.flags and FLAG_SYSTEM != 0
+    val ignoreAllowBackup = !isSystemApp
+        && (context.checkSelfPermission(Manifest.permission.LOG_COMPAT_CHANGE)
+        == PackageManager.PERMISSION_GRANTED)
+        && (context.checkSelfPermission(Manifest.permission.READ_COMPAT_CHANGE_CONFIG)
+        == PackageManager.PERMISSION_GRANTED)
+        && CompatChanges.isChangeEnabled(IGNORE_ALLOW_BACKUP_IN_D2D, packageName,
+        UserHandle.of(userId))
```
grote, line 171: Under what circumstances can this `IGNORE_ALLOW_BACKUP_IN_D2D` be false?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -50,17 +53,14 @@ internal class PackageService(
                 }
             }
 
-            val eligibleApps =
-                backupManager.filterAppsEligibleForBackupForUser(myUserId, packages.toTypedArray())
```
grote, line 54: Why not leave it to the system to tell us which apps are eligible?

app/src/main/java/com/stevesoltys/seedvault/metadata/Metadata.kt
Uldiniad: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/backup/java/com/android/server/backup/restore/ActiveRestoreSession.java;l=302?q=getBackupEligibilityRules

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
Uldiniad: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java;l=82;drc=dada247f224cfd53ab6b5a56c2bd136badea12f0;bpv=0;bpt=1
It cannot be imported

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
Uldiniad: It returns the wrong list.
https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/backup/java/com/android/server/backup/UserBackupManagerService.java;l=585;drc=dada247f224cfd53ab6b5a56c2bd136badea12f0;bpv=0;bpt=1 uses OperationType.BACKUP. We want OperationType.MIGRATION

I'm doing the necessary checks in BackupCoordinator instead

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
Uldiniad: https://review.calyxos.org/c/CalyxOS/platform_frameworks_base/+/10624/1/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
Uldiniad: https://github.com/seedvault-app/seedvault/pull/407/files/396d5792b8f31ecc8bf175f3f54a14930748dc9f#r875738119 see my reply

grote:
The ticket that this is resolving is https://github.com/seedvault-app/seedvault/issues/165 which has a more appropriate title as we are not actually doing d2d transfers here, but faking to be one.

app/src/main/java/com/stevesoltys/seedvault/metadata/Metadata.kt
chirayudesai: > TODO(b/182986784): Remove device name comparison once a designated field for operation type is added to RestoreSet object.

That sounds like it's related to https://github.com/seedvault-app/seedvault/issues/165#issuecomment-938045174

grote: This has a TODO that they plan to do this differently. Maybe we are too early with this.

> We should make the UI portion use ${Build.MANUFACTURER} ${Build.MODEL}

How about not changing the metadata info at all, but setting the restore set name/device to D2D instead?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
grote: Maybe another option around this could be to take the scheduling into our own hands and then use the methods that allow us to pass the operation type.

Doing the scheduling ourselves has other advantages, e.g. offline, metered networks, usb, etc. which are already tricky with AOSPs scheduling. But it would be way more work as well.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
grote: If there's really no way around copying that logic from AOSP, then please let's do it in a similarly readable fashion.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
grote: Could we then also add the docs that AOSP has for this that explain what this constant does?
A link to where it comes from would also be nice.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
chirayudesai: Note that the above change was just to make testing easier, not planning to merge.

@Uldiniad when you move this to a method, it could be a comment.

grote: It would be nice if you could write a bit more prose instead just pasting links in the hope I can figure it out myself.

So what does this code mean? Is this something per-app? depending on the app's target sdk version? is it enabled automatically per app as soon as they upgrade their target?

chirayudesai: > Is this something per-app? 

Yes.

> depending on the app's target sdk version?

Yes, AOSP has it set to target SDK S and above

> is it enabled automatically per app as soon as they upgrade their target?

Probably? Given that it checks the apps target sdk version / eligibility when backing up.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
grote: I don't see how those rules get used? Are they not asking the transport anymore what it considers eligible?

How certain are you that there is no other way to call into system methods with a different operation type so we get the OS rules without having to duplicate them ourselves? E.g. can we call `BackupManagerService.getEligibilityRulesForOperation(operationType)`?

Do you know what happens if we try to back up an app that AOSP doesn't consider eligible even when using `OperationType.MIGRATION`?

grote:
When testing this, please pay special attention to the backup notification progress bar and success message with the app total.

app/src/main/java/com/stevesoltys/seedvault/metadata/Metadata.kt
Uldiniad: Sounds good to me

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
Uldiniad: Concerning scheduling, I am unsure that can be done without OS changes.

`getEligibilityRulesForOperation` is not exposed in current public nor system APIs. Without OS modifications we cannot make use of it. The ideal for most compatibility for this feature is to have no OS changes required. A number of calls to determine eligibility are hardcoded in the system.

Concerning app eligibility for backup, the only difference is that `OperationType.BACKUP` only checks `allowBackup` while `OperationType.MIGRATION` also checks first if `allowBackup` can be ignored. So there would be no case where the list of eligible apps returned by `OperationType.BACKUP` contains apps that are not present in  `OperationType.MIGRATION` https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/backup/java/com/android/server/backup/utils/BackupEligibilityRules.java;l=163;drc=4875b203ceba30d2d50d642da04064ca37b66e43;bpv=0;bpt=1

Uldiniad:
> 

Have done. Everything looks like I would expect. Apps that do not support D2D still show under the "could not backup app" section

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
t-m-w: When testing this PR, an exclusion for DAVx5 did not take effect for me. I reverted this section to the previous eligibleApps code (also bringing back backupManager to the initializer, etc.), and the exclusion took effect. DAVx5 was the only app I attempted to exclude, but I would guess that other apps would have been affected as well.

[EDIT: ...On the other hand, reverting to the previous behavior seems to exclude D2D apps anyway, so that's kind of pointless.]

Uldiniad: Right, I had disabled this code due to it removing packages that are eligible under D2D. I am not sure if the handling for not backing up the app handling the backup, should be done here or if it can/should be done in `isAppEligibleForBackup`. I would defer to @grote on this

t-m-w: These changes, derived from your commented code, have been working for me currently. The backup-handling app seems to be successfully addressed elsewhere. https://gitlab.com/tmw_calyxos_fork/platform_packages_apps_Seedvault/-/commit/96e049b80a62fbd2bcd50d3b6b772ba169c087d0

t-m-w:
[EDIT: Still some issues to address related to the backup/restore provider app. I just started testing a large restore with DAVx5 and everything imploded after it restored the apk for DAVx5. (issue #374)]

With some [changes to fix app exclusions](https://gitlab.com/tmw_calyxos_fork/platform_packages_apps_Seedvault/-/commit/96e049b80a62fbd2bcd50d3b6b772ba169c087d0), this seems to work well! It is able to back up and restore Bromite, for example, allowing the user to keep their bookmarks, homepage, and other settings around. (It would probably fail due to quota if I actually browsed very much in Bromite, first, but I'm not sure.)

One interesting thing is that it tries to back up Android System Webview, but it never succeeds at that (possibly due to `Cancel full backup of com.android.webview because of NO_DATA`?), instead having an exclamation point in an orange circle. As mentioned earlier, users may find it useful to have their bookmarks and browser settings backed up, so it would be nice if this could affect Chromium, but that might be far out of scope.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
t-m-w: Okay, I may have to retract that -- I just started testing a large restore with DAVx5 and everything imploded after it restored the apk for DAVx5. (issue #374)

chirayudesai:
Friendly bump :)

@grote @Uldiniad where are we on this? ~~Needs rebasing for 13.~~

Edit: It rebased cleanly, untested.

PR #412, opened 2022-05-31
Tell user about non enabled backup

### LINKED ISSUE #390: Notify user if they have not selected anything to back up
I was trying to help a user set up seedvault.  They enabled backups.  They verified the 12 words.  They hit 'backup now' and nothing happened.   They said hey it doesn't work.

The reason nothing happened was because they had not selected either 'backup my apps' or 'backup my files'.   But it would have been helpful if seedvault had said  something like 'You have not selected either backup apps or backup files. '

BadDaemon:
* Pressing "Backup now" should not just ignore you when it's not actually
  enabled
* Add a toast message telling you it's not enabled

Fixes: #390
Change-Id: I5d698a244652e094b0acfc42cdea017a0af9a20b

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -111,6 +111,7 @@
     <string name="notification_channel_title">Backup notification</string>
     <string name="notification_title">Backup running</string>
     <string name="notification_backup_already_running">Backup already in progress</string>
+    <string name="notification_backup_disabled">Backup not enabled</string>
 
     <string name="notification_success_title">Backup finished</string>
     <string name="notification_success_text">%1$d of %2$d apps backed up. Tap to learn more.</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -111,6 +111,7 @@
     <string name="notification_channel_title">Backup notification</string>
     <string name="notification_title">Backup running</string>
     <string name="notification_backup_already_running">Backup already in progress</string>
+    <string name="notification_backup_disabled">Backup not enabled</string>
```
chirayudesai, line 114: Maybe add something about it not enabled at the OS level, so that the user has a hint about possible solutions

chirayudesai: D'oh

chirayudesai (approved):

PR #414, opened 2022-06-06
Mark background backup notification as ongoing

joserebelo:
Background backup notifications are not marked as ongoing. This will make them get propagated to external devices in very quick succession on every update (eg. [by Gadgetbridge](https://codeberg.org/Freeyourgadget/Gadgetbridge/issues/2611))

I can't currently test this (have Seedvault as a system app), but it's a simple enough change - unless there was some specific reason the notification is not marked as ongoing?


chirayudesai (approved):
I do have Gadgetbridge set to display Seedvault notifications here and for the most part it handles them fine, but this looks correct to me either way.

Thanks!

PR #422, opened 2022-06-30
Exclude storage provider from restore

t-m-w:
Excludes the storage provider package from the restoration process. Otherwise, when the storage provider (i.e. DAVx5 or Nextcloud) is installed, it will be terminated, and the restoration process will be interrupted.

Should fix issue #374.

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
@@ -41,7 +41,12 @@ internal class ApkRestore(
     @Suppress("BlockingMethodInNonBlockingContext")
     fun restore(backup: RestorableBackup) = flow {
         // filter out packages without APK and get total
-        val packages = backup.packageMetadataMap.filter { it.value.hasApk() }
+        val packages = backup.packageMetadataMap.filter {
+            // We also need to exclude the DocumentsProvider used to retrieve backup data.
+            // Otherwise, it gets killed when we install it, terminating our restoration.
+            val isStorageProvider = it.key == storagePlugin.providerPackageName
+            it.value.hasApk() && !isStorageProvider
+        }
         val total = packages.size
         var progress = 0
 
--- a/app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkBackupRestoreTest.kt
+++ b/app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkBackupRestoreTest.kt
@@ -118,6 +118,7 @@ internal class ApkBackupRestoreTest : TransportTest() {
         every { metadataManager.salt } returns salt
         every { crypto.getNameForApk(salt, packageName) } returns name
         every { crypto.getNameForApk(salt, packageName, splitName) } returns suffixName
+        every { storagePlugin.providerPackageName } returns storageProviderPackageName
 
         apkBackup.backupApkIfNecessary(packageInfo, PackageState.APK_AND_DATA, outputStreamGetter)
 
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
```diff
@@ -40,8 +40,13 @@ internal class ApkRestore(
...
     @Suppress("BlockingMethodInNonBlockingContext")
     fun restore(backup: RestorableBackup) = flow {
-        // filter out packages without APK and get total
-        val packages = backup.packageMetadataMap.filter { it.value.hasApk() }
+        // filter out packages without APK, along with the storage provider, and get total
+        val packages = backup.packageMetadataMap.filter {
+            it.value.hasApk() &&
+            // We need to exclude the DocumentsProvider used to retrieve backup data.
+            // Otherwise, it gets killed when we install it, terminating our restoration.
+            it.key != storagePlugin.providerPackageName
```
grote: Nitpick: I find this a bit hard to read. How about introducting a val here: `val isStorageProvider = it.key == storagePlugin.providerPackageName` and adding the comments there and then have the filter condition in a single line?

grote (commented):
Thanks good catch! This looks good, just one nitpick here.

As people sometimes switch around backup location providers manually, it probably makes sense to filter here and not when backing up.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
t-m-w: All right, I attempted to make that adjustment and some minor comments tweaks in the latest commit (is that how I'm supposed to add new PR changes? I'm new to this); if I misinterpreted the requested change, please let me know.

I should say that those comments were copied and pasted from [BackupCoordinator.kt](https://github.com/seedvault-app/seedvault/blob/976e898656a59eb9cc5a7b4d0a0f8e2a978fc875/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt#L149-L150) with minor word changes.

grote: Thanks, looks good now. However, there's 11 tests failing now. These need to be fixed first and then ideally add a test that ensures that the `storagePlugin.providerPackageName` gets excluded.

You can squash your commits and force push to your branch to keep logical changes together in a single commit.

t-m-w: I'd like to add the test you've described, but unit tests are new to me, so first I want to make sure that I've correctly fixed the failing tests. I force-pushed an adjustment that adds relaxed=true to StoragePlugin. Is this too...relaxed of a way to handle the fact that providerPackageName does not return a value in the tests?

app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkBackupRestoreTest.kt
```diff
@@ -49,7 +49,7 @@ internal class ApkBackupRestoreTest : TransportTest() {
 
     @Suppress("Deprecation")
     private val legacyStoragePlugin: LegacyStoragePlugin = mockk()
-    private val storagePlugin: StoragePlugin = mockk()
+    private val storagePlugin: StoragePlugin = mockk(relaxed = true)
```
grote: That *does* fix the test, but I think it would be better to add a return value explicitly where needed:

example:

`every { storagePlugin.providerPackageName } returns "org.example"`

This way, you can also add a new test that checks that if the returned package name would normally be restored, it won't get restored if returned by the storagePlugin.

t-m-w:
OK, here's another go at it.

I'm not sure how you'll feel about my adding of the value `storageProviderPackageName` in TransportTest.kt, but specifying the same literal string repeatedly felt a bit off. It seems to work with returning "", which is lower-maintenance, but could cause problems in the future.

Also, the test function I added includes only enough mocked values to pass the test. If I adjust it such that `storagePlugin.providerPackageName` returns some other value, like "aaa", it appropriately fails, but with `io.mockk.MockKException: no answer found` for various missing mocked values. Is this an acceptable way to fail?

grote (approved):
Thanks for improving the tests and sorry for taking so long for review!

Issue #433, opened 2022-08-19
Port to Android 13


grote:
Branch: https://github.com/seedvault-app/seedvault/tree/android13

Also check upstream commits for any changes to the backup related code.

grote:
Some behavior changes to check:

* [x] Apps need to ask permission to show notifications - https://developer.android.com/about/versions/13/changes/notification-permission
* [ ] New rules for intent filters could affect our use of intents to open other apps  https://developer.android.com/about/versions/13/behavior-changes-13#security
* [x] must request more granular media permissions instead of the `READ_EXTERNAL_STORAGE` https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions

grote:
* There's an [`EncryptedLocalTransport`](https://android.googlesource.com/platform/frameworks/base.git/+/android-13.0.0_r3/packages/EncryptedLocalTransport) now. Not sure if functional.
* But looks like the "Encryption in the Framework project" [is dead](https://android.googlesource.com/platform/frameworks/base.git/+/5169896db8741a7d1cadefb3616d8f1ed604fef6)? 
* timeout in `TransportStatusCallback` was [halfed to 5min](https://android.googlesource.com/platform/frameworks/base.git/+/3f82689d2b59f1437479ccb3fb69e662a98ac2fb%5E%21/#F1) which hopefully doesn't cause issues when we do stuff like long-running APK backups during a package backup 
* Allow BackupHelper [to be overridden](https://android.googlesource.com/platform/frameworks/base.git/+/8de1c313e64c40f48b222e6cb069a4509723b95c%5E%21/#F0). This will allow vendor versions of System UI to backup their vendor-specific data.

Also lots of internal bug fixes which might solve some issues we've seen.

chirayudesai:
> `EncryptedLocalTranspot`

The commit adding it has some more info: https://android.googlesource.com/platform/frameworks/base.git/+/004e85f798dc88265edffebeba66da4efaeb2735 

CTS tests are public, GTS not.

> "Encryption in the Framework"

Linked commit mentions "making the backup transport AIDL async"

> Transport timeout

I guess we could always consider optionally bumping it - it'd be a quality of life patch that ROMs could apply if they want

> vendor versions of System UI 

Google ships SystemUIGoogle on Pixels for example (where a LOT of features are implemented)

chirayudesai:
[...]

PR #441, opened 2022-09-08
Handle null better for Android 13

grote:
Though we noticed this only on Android 13, this MR *could* probably also be backported to 12 if somebody cared.

Background info starting here: https://github.com/seedvault-app/seedvault/issues/433#issuecomment-1239483667

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
@@ -381,7 +381,13 @@ internal class BackupCoordinator(
                 }
                 // hook in here to back up APKs of apps that are otherwise not allowed for backup
                 if (isPmBackup && settingsManager.canDoBackupNow()) {
-                    backUpApksOfNotBackedUpPackages()
+                    try {
+                        backUpApksOfNotBackedUpPackages()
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
```diff
@@ -381,7 +381,13 @@ internal class BackupCoordinator(
                 }
                 // hook in here to back up APKs of apps that are otherwise not allowed for backup
                 if (isPmBackup && settingsManager.canDoBackupNow()) {
-                    backUpApksOfNotBackedUpPackages()
+                    try {
+                        backUpApksOfNotBackedUpPackages()
+                    } catch (e: Exception) {
+                        Log.e(TAG, "Error backing up APKs of opt-out apps: ", e)
+                        // We are re-throwing this, because we want to know about problems here
+                        throw e
```
grote, line 389: Note that `IOException`s are already caught on a per-app basis inside `backUpApksOfNotBackedUpPackages()`.

chirayudesai:
Did a quick restore of a 12L backup to a 13 device with this, went ok.

Edit: Subsequent backup worked too.

theimpulson (approved):

PR #446, opened 2022-09-13
RecoveryCode: Allow screenshots on debug builds

chirayudesai:
* We already pre-fill the code on debug builds to make testing easier
* Allow screenshots too so we don't have to note down the code right away when testing

Change-Id: Ie9cc3080616b78a6e5f9b427c080f10c146a036d

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/recoverycode/RecoveryCodeActivity.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/recoverycode/RecoveryCodeActivity.kt
@@ -4,6 +4,7 @@ import android.os.Bundle
 import android.view.MenuItem
 import android.view.WindowManager.LayoutParams.FLAG_SECURE
 import com.stevesoltys.seedvault.R
+import com.stevesoltys.seedvault.isDebugBuild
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/stevesoltys/seedvault/ui/recoverycode/RecoveryCodeActivity.kt
```diff
@@ -15,7 +15,7 @@ class RecoveryCodeActivity : BackupActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        window.addFlags(FLAG_SECURE)
+        if (!isDebugBuild()) window.addFlags(FLAG_SECURE)
```
grote, line 19: CI failed. Maybe you need to import this?

grote (changes requested):

app/src/main/java/com/stevesoltys/seedvault/ui/recoverycode/RecoveryCodeActivity.kt
chirayudesai: Done.

grote (approved):

PR #458, opened 2022-10-14
Mark Nextcloud as not recommended

chirayudesai:
* We've had a love/hate relationship with the combination of
  SAF + Nextcloud app + Nextcloud server
* SAF is quite limiting, and that causes it's own issues
* On top of that, Nextcloud app has it's own issues. We've
  even sent patches for some things, but there's only so much one can do
  https://github.com/nextcloud/android/pulls?q=is%3Apr+author%3Agrote
* With the DAVx5 WebDAV support, it seems to work better.
* However, most recently a server side change broke things entirely, with either apps
  https://github.com/seedvault-app/seedvault/issues/490#issuecomment-1368032414
* Due of all of this, it just leads to a bad backup experience.
  As such, let's just say this is not recommended at all entirely
* Hopefully in the future we'll be able to support more options


Popkornium18:
Out of curiosity: Why is Nextcloud not recommended?

grote:
> Out of curiosity: Why is Nextcloud not recommended?

We've seen lots of issues with their app, fixed some ourselves, but found the DavX to perform much better in practice.

chirayudesai:
> 1. Add detailed rationale

Done

> 2. Dedicated string that can be translated

Also done, not sure if this is the best way.

chirayudesai:
Screenshot_20230106-044000

app/src/main/res/values/strings.xml
```diff
@@ -59,6 +59,7 @@
     <string name="storage_fake_drive_summary">Needs to be plugged in</string>
     <string name="storage_available_bytes"><xliff:g example="1 GB" id="size">%1$s</xliff:g> free</string>
     <string name="storage_fake_nextcloud_title" translatable="false">Nextcloud</string>
+    <string name="not_recommended">Not recommended</string>
```
grote: can we not add the brackets here and simply concatenate the two strings?

t-m-w: Regrettably, I'd probably be the last to know for sure whether this is the case, but it seems to me like either way of doing this may pose localization challenges.

What about something like:
`<string name="not_recommended_provider"><xliff:g example="Skynet">%1$s</xliff:g> (Not recommended)</string>`

...and plugging in whatever the provider is? This way, translators can figure out the proper way of wording the parenthetical and where it should go.

chirayudesai:
Done.

chirayudesai:
I considered this but is it really worth it?

mikeNG:
> I considered this but is it really worth it?

That's the correct way

app/src/main/res/values/strings.xml
chirayudesai: Done.

chirayudesai:
> > I considered this but is it really worth it?
> 
> That's the correct way

Done.

grote (approved):

PR #465, opened 2022-11-11
Add expert option to save logs

### LINKED ISSUE #464: Support sharing logs directly

grote:
This saves the log to a file (way easier than sharing a file without storing it)

Screenshot_20221111_173844

User gets the choice to also store system logs

Screenshot_20221111_173449

Closes #464

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/settings/ExpertSettingsFragment.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/settings/ExpertSettingsFragment.kt
@@ -1,14 +1,34 @@
 package com.stevesoltys.seedvault.settings
 
 import android.os.Bundle
+import androidx.activity.result.contract.ActivityResultContracts.CreateDocument
+import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
 import com.stevesoltys.seedvault.R
 import com.stevesoltys.seedvault.permitDiskReads
+import com.stevesoltys.seedvault.transport.backup.PackageService
+import org.koin.android.ext.android.inject
+import org.koin.androidx.viewmodel.ext.android.sharedViewModel
 
 class ExpertSettingsFragment : PreferenceFragmentCompat() {
+
+    private val viewModel: SettingsViewModel by sharedViewModel()
+    private val packageService: PackageService by inject()
+    // TODO set mimeType when upgrading androidx lib
+    private val createFileLauncher = registerForActivityResult(CreateDocument()) { uri ->
+        viewModel.onLogcatUriReceived(uri)
+    }
+
     override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
         permitDiskReads {
             setPreferencesFromResource(R.xml.settings_expert, rootKey)
+        }
+        findPreference<Preference>("logcat")?.setOnPreferenceClickListener {
+            val versionName = packageService.getVersionName(requireContext().packageName) ?: "ver"
+            val timestamp = System.currentTimeMillis()
+            val name = "seedvault-$versionName-$timestamp.txt"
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/stevesoltys/seedvault/settings/ExpertSettingsFragment.kt
```diff
@@ -1,15 +1,29 @@
...
+    private val createFileLauncher = registerForActivityResult(CreateDocument()) { uri ->
+        viewModel.onLogcatUriReceived(uri)
+    }
+
     override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
         permitDiskReads {
             setPreferencesFromResource(R.xml.settings_expert, rootKey)
         }
+        findPreference<Preference>("logcat")?.setOnPreferenceClickListener {
+            createFileLauncher.launch("seedvault-log.txt")
```
chirayudesai: Let's add some more info here, perhaps version and timestamp, so one can have multiple logs.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsFragment.kt
```diff
@@ -105,7 +105,7 @@ class SettingsFragment : PreferenceFragmentCompat() {
                     dialog.dismiss()
                 }
                 .setNegativeButton(R.string.settings_backup_apk_dialog_disable) { dialog, _ ->
-                    apkBackup.isChecked = enable
+                    apkBackup.isChecked = false
```
chirayudesai, line 108: Why this change? It does make sense.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
```diff
@@ -233,4 +237,27 @@ internal class SettingsViewModel(
         BackupJobService.cancelJob(app)
     }
 
+    fun onLogcatUriReceived(uri: Uri?) = viewModelScope.launch(Dispatchers.IO) {
+        if (uri == null) {
+            onLogcatError()
+            return@launch
+        }
+        val command = "logcat -d --uid=1000,${myUid()} *:V"
```
chirayudesai, line 246: maybe add a comment here
// 1000 is system uid, needed to get backup logs from the OS code.

chirayudesai (commented):
Minor nits, LGTM.

Will smoke test before merge

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsFragment.kt
grote: The linter suggested that, because `enable` is always `false` here.

chirayudesai:
I hate to be pedantic but I'd also like to have proper commits, given we do that across all other projects.

Right now we have 3 commits here,
1 adds logs and makes some unrelated changes, and 2 and 3 are followups.

It'd be much better if it was:
1 add logs (and the followups could be squashed here, fine either way)
2 the minor unrelated changes.

It's ok to have those minor changes in the same PR, but at the very least they should be separate commits to make it easier to back out a change if needed, and also just git blame / browsing commits.

We could also just make those changes across the entire repository later rather than doing it file by file like this when working on other things to avoid unrelated changes.

grote:
reordered commits as requested

chirayudesai (approved):

chirayudesai:
> reordered commits as requested

Thanks, somehow I missed this.

I pinged @mikeNG about the strings, apart from that I'll do a quick smoke test and merge, LGTM.

theimpulson (approved):

mikeNG (approved):
Strings look fine to me

PR #473, opened 2022-11-17
Pretend to be a device to device transfer - fake-d2d

Uldiniad:
Previous PR: https://github.com/seedvault-app/seedvault/pull/407

Issue: https://github.com/seedvault-app/seedvault/issues/165

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
@@ -5,7 +5,10 @@ import android.hardware.usb.UsbDevice
 import android.net.ConnectivityManager
 import android.net.NetworkCapabilities
 import android.net.Uri
+import android.util.Log
 import androidx.annotation.UiThread
+import androidx.annotation.VisibleForTesting
+import androidx.annotation.VisibleForTesting.PRIVATE
 import androidx.annotation.WorkerThread
 import androidx.documentfile.provider.DocumentFile
 import androidx.preference.PreferenceManager
@@ -34,6 +37,8 @@ private const val PREF_KEY_UNLIMITED_QUOTA = "unlimited_quota"
 
 class SettingsManager(private val context: Context) {
 
+    private val TAG = SettingsManager::class.java.simpleName
+
     private val prefs = permitDiskReads {
         PreferenceManager.getDefaultSharedPreferences(context)
     }
@@ -42,11 +47,18 @@ class SettingsManager(private val context: Context) {
     private var token: Long? = null
 
     /**
+     * The StoragePlugin provider package name is assigned by ConfigurableBackupTransportService
+     * to ensure it is excluded; if the provider is killed, the backup or restore will fail.
+     */
+    var pluginProviderPackageName: String? = null
+
+    /**
      * This gets accessed by non-UI threads when saving with [PreferenceManager]
      * and when [isBackupEnabled] is called during a backup run.
      * Therefore, it is implemented with a thread-safe [ConcurrentSkipListSet].
      */
-    private val blacklistedApps: MutableSet<String> by lazy {
+    @VisibleForTesting(otherwise = PRIVATE)
+    val blacklistedApps: MutableSet<String> by lazy {
         ConcurrentSkipListSet(prefs.getStringSet(PREF_KEY_BACKUP_APP_BLACKLIST, emptySet()))
     }
 
@@ -131,6 +143,15 @@ class SettingsManager(private val context: Context) {
     }
 
     fun isBackupEnabled(packageName: String) = !blacklistedApps.contains(packageName)
+
+    fun isAppAllowedForBackup(packageName: String): Boolean {
+        // Check that the app is not blacklisted by the user
+        val enabled = isBackupEnabled(packageName)
+        if (!enabled) Log.w(TAG, "Backup of $packageName disabled by user.")
+        // We need to exclude the DocumentsProvider used to store backup data.
+        // Otherwise, it gets killed when we back it up, terminating our backup.
+        return enabled && packageName != pluginProviderPackageName
+    }
 
     fun isStorageBackupEnabled() = prefs.getBoolean(PREF_KEY_BACKUP_STORAGE, false)
 
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/ConfigurableBackupTransport.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/ConfigurableBackupTransport.kt
@@ -1,6 +1,7 @@
 package com.stevesoltys.seedvault.transport
 
 import android.app.backup.BackupAgent.FLAG_CLIENT_SIDE_ENCRYPTION_ENABLED
+import android.app.backup.BackupAgent.FLAG_DEVICE_TO_DEVICE_TRANSFER
 import android.app.backup.BackupTransport
 import android.app.backup.RestoreDescription
 import android.app.backup.RestoreSet
@@ -20,7 +21,9 @@ import org.koin.core.component.inject
 // If we ever change this, we should use a ComponentName like the other backup transports.
 val TRANSPORT_ID: String = ConfigurableBackupTransport::class.java.name
 
-const val TRANSPORT_FLAGS = FLAG_CLIENT_SIDE_ENCRYPTION_ENABLED
+// Since there seems to be consensus in the community to pose as device to device transport,
+// we are pretending to be one here. This will backup opt-out apps that target API 30.
+const val TRANSPORT_FLAGS = FLAG_CLIENT_SIDE_ENCRYPTION_ENABLED or FLAG_DEVICE_TO_DEVICE_TRANSFER
 
 private const val TRANSPORT_DIRECTORY_NAME =
     "com.stevesoltys.seedvault.transport.ConfigurableBackupTransport"
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/ConfigurableBackupTransport.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/ConfigurableBackupTransport.kt
@@ -12,6 +12,7 @@ import android.os.ParcelFileDescriptor
 import android.util.Log
 import com.stevesoltys.seedvault.R
 import com.stevesoltys.seedvault.settings.SettingsActivity
+import com.stevesoltys.seedvault.settings.SettingsManager
 import com.stevesoltys.seedvault.transport.backup.BackupCoordinator
 import com.stevesoltys.seedvault.transport.restore.RestoreCoordinator
 import kotlinx.coroutines.runBlocking
@@ -22,7 +23,7 @@ import org.koin.core.component.inject
 val TRANSPORT_ID: String = ConfigurableBackupTransport::class.java.name
 
 // Since there seems to be consensus in the community to pose as device to device transport,
-// we are pretending to be one here. This will backup opt-out apps that target API 30.
+// we are pretending to be one here. This will backup opt-out apps that target at least API 30.
 const val TRANSPORT_FLAGS = FLAG_CLIENT_SIDE_ENCRYPTION_ENABLED or FLAG_DEVICE_TO_DEVICE_TRANSFER
 
 private const val TRANSPORT_DIRECTORY_NAME =
@@ -38,6 +39,7 @@ class ConfigurableBackupTransport internal constructor(private val context: Cont
 
     private val backupCoordinator by inject<BackupCoordinator>()
     private val restoreCoordinator by inject<RestoreCoordinator>()
+    private val settingsManager by inject<SettingsManager>()
 
     override fun transportDirName(): String {
         return TRANSPORT_DIRECTORY_NAME
@@ -120,9 +122,9 @@ class ConfigurableBackupTransport internal constructor(private val context: Cont
 
     override fun isAppEligibleForBackup(
         targetPackage: PackageInfo,
-        isFullBackup: Boolean,
+        @Suppress("UNUSED_PARAMETER") isFullBackup: Boolean,
     ): Boolean {
-        return backupCoordinator.isAppEligibleForBackup(targetPackage, isFullBackup)
+        return settingsManager.isAppAllowedForBackup(targetPackage.packageName)
     }
 
     override fun getBackupQuota(packageName: String, isFullBackup: Boolean): Long = runBlocking {
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
@@ -1,6 +1,5 @@
 package com.stevesoltys.seedvault.transport.backup
 
-import android.app.backup.IBackupManager
 import android.content.Context
 import android.content.pm.ApplicationInfo.FLAG_ALLOW_BACKUP
 import android.content.pm.ApplicationInfo.FLAG_STOPPED
@@ -17,6 +16,7 @@ import android.util.Log
 import android.util.Log.INFO
 import androidx.annotation.WorkerThread
 import com.stevesoltys.seedvault.MAGIC_PACKAGE_MANAGER
+import com.stevesoltys.seedvault.settings.SettingsManager
 
 private val TAG = PackageService::class.java.simpleName
 
@@ -28,13 +28,13 @@ private const val LOG_MAX_PACKAGES = 100
  */
 internal class PackageService(
     private val context: Context,
-    private val backupManager: IBackupManager,
+    private val settingsManager: SettingsManager,
 ) {
 
     private val packageManager: PackageManager = context.packageManager
     private val myUserId = UserHandle.myUserId()
 
-    val eligiblePackages: Array<String>
+    val requestedPackages: Array<String>
         @WorkerThread
         @Throws(RemoteException::class)
         get() {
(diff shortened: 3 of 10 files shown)


chirayudesai:
@Uldiniad compared to the latest commits on the previous PR https://github.com/seedvault-app/seedvault/pull/407 (which were on top of android13 even if the branch was called android12) - have you made any other changes here yet?

Uldiniad:
Yes. The only thing in common with the previous patch is the first commit. Second commit is a complete overhaul made by @t-m-w

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -159,7 +172,12 @@ internal fun PackageInfo.isSystemApp(): Boolean {
 
 internal fun PackageInfo.allowsBackup(): Boolean {
     if (packageName == MAGIC_PACKAGE_MANAGER || applicationInfo == null) return false
-    return applicationInfo.flags and FLAG_ALLOW_BACKUP != 0
+
+    // At backup time, the system will filter out any apps that *it* does not want to be backed up.
+    // Now that we have switched to D2D, *we* generally want to back up as much as possible;
+    // part of the point of D2D is to ignore FLAG_ALLOW_BACKUP (allowsBackup). So, we return true.
+    // See Google Change ID 183147249.
```
chirayudesai: What does this name?

That looks like a bug number, or is that a gerrit change? Link would be better.

chirayudesai (commented):
Minor nit. Low hove small the change is now :D

t-m-w:
This has been tested with backup sets created by Seedvault with and without D2D support , and restores using Seedvault with and without D2D support. If the Seedvault app is older (does not support D2D), any app whose data backup/restore requires D2D (for example, Bromite) will simply not have its data restored. Unfortunately, there is no error message for the app whose data failed to restore in this situation -- merely a red exclamation mark -- but this is part of wider error-handling issues in Seedvault.

One of the differences between this PR and #407 is that the full eligibility determination is largely left up to the system at backup time, rather than the logic being replicated in Seedvault; rather, Seedvault just ensures that *excluded* apps are not sent to the backup manager. The comments in the code indicate this.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
t-m-w: It's referenced in Google's code. This is the only place I can find it other than that, though: https://developer.android.com/about/versions/12/reference/compat-framework-changes

chirayudesai: Ah then it's likely on their internal Gerrit

app/src/main/java/com/stevesoltys/seedvault/transport/ConfigurableBackupTransport.kt
```diff
@@ -20,7 +21,9 @@ import org.koin.core.component.inject
 // If we ever change this, we should use a ComponentName like the other backup transports.
 val TRANSPORT_ID: String = ConfigurableBackupTransport::class.java.name
 
-const val TRANSPORT_FLAGS = FLAG_CLIENT_SIDE_ENCRYPTION_ENABLED
+// Since there seems to be consensus in the community to pose as device to device transport,
+// we are pretending to be one here. This will backup opt-out apps that target API 30.
```
grote: nitpick: target *at least* API 30?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -29,29 +32,39 @@ private const val LOG_MAX_PACKAGES = 100
 internal class PackageService(
     private val context: Context,
     private val backupManager: IBackupManager,
-) {
+) : KoinComponent {
 
     private val packageManager: PackageManager = context.packageManager
     private val myUserId = UserHandle.myUserId()
+    private val backupCoordinator by inject<BackupCoordinator>()
```
grote: Instead of injecting the `BackupCoordinator` here in a cyclic dependency (which will blow up once we consistently use an injection framework such as Hilt), could we move the `isAppEligibleForBackup` code into another place where it then gets used by `PackageService` and `BackupCoordinator`? Maybe `SettingsManager` would work?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -29,29 +32,39 @@ private const val LOG_MAX_PACKAGES = 100
...
+            // apps; it is not designed to determine MIGRATION eligibility, only eligibility for
+            // inclusion in *scheduled* BACKUPs, which are implicitly *not* D2D migrations.
+            // None of the other eligibility methods are exposed by AOSP APIs. On the other hand,
+            // the actual backup process properly utilizes OperationType.MIGRATION and performs its
+            // own checks as to whether apps are allowed to be backed up. All we need to do now is
+            // filter out apps that *we* want to be excluded. The system will do the rest later.
+            val eligibleApps = packages.filter {
+                val isFullBackup = if (it.applicationInfo.backupAgentName != null)
+                    it.applicationInfo.flags and FLAG_FULL_BACKUP_ONLY != 0
+                    else true
```
grote: Could you reformat this to be a bit more readable? Actually scratch that. We don't even care about is `isFullBackup`. So if we move the `isAppEligibleForBackup()` into another place, we could just remove the `isFullBackup` parameter.

grote (changes requested):
Very nice! Much better than the original PR. Bonus points for avoiding to mess too much with PackageService.

Just left some small nitpicks.

I currently don't have time for own testing, so I trust that this will work.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
t-m-w: Okay, a few issues, based on a work in progress:
1. `SettingsManager` does not have access to the current storage plugin, which `isAppEligibleForBackup` needs in order to exclude the storage plugin's package name. I've decided to try providing the `plugin` to `SettingsManager`. I'm not sure if that's acceptable.
2. Moving `isAppEligibleForBackup` to `SettingsManager` also means that the test needs to be updated, likely moved out of `BackupCoordinatorTest` and into `SettingsManagerTest`, which does not currently exist. I've started creating it.
3. The existing test relies on `packageInfo`, which is defined in `TransportTest`. It feels possibly odd to have `SettingsManagerTest` test inherit from `TransportTest`, but I can either try that or I can re-define some test package info in `SettingsManagerTest`. The former is nice to prevent code duplication, but maybe there's some better way.

Separately from all of this, I'd like to rename `eligiblePackages` and `isAppEligibleForBackup` to `requestedPackages` and `isAppRequestedForBackup`, respectively, to better reflect their new roles and make it a little less confusing.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
t-m-w: Done

app/src/main/java/com/stevesoltys/seedvault/transport/ConfigurableBackupTransport.kt
t-m-w: Done

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
t-m-w: ...Done, but the patch is much bigger now, unfortunately. This wasn't easy. I don't mind undoing it if needed. Learned a little more about mockk along the way.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
t-m-w: Done; it's all moved to `SettingsManager` now.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
```diff
@@ -34,19 +37,28 @@ private const val PREF_KEY_UNLIMITED_QUOTA = "unlimited_quota"
...
 
     @Volatile
     private var token: Long? = null
 
+    /**
+     * The StoragePlugin provider package name is assigned by ConfigurableBackupTransportService
+     * to ensure it is excluded; if the provider is killed, the backup or restore will fail.
+     */
+    var pluginProviderPackageName: String? = null
+
```
grote, line 54: Sorry, but this won't fly. Making the setter of an essential property publicly accessible isn't good encapsulation and worse you need to rely on whatever external component to set this at the right time. Can you be sure this always happens when it needs to?

The doc above is nice and all, but I'd prefer robust code instead. For example, in `ConfigurableBackupTransportService`, this is only called in the `requestBackup()` method. Is this enough? How can you be sure? AFAIK, this method is only used when the user triggers a backup manually, but not when the system runs a backup. You can use `adb shell bmgr` for testing this.

So let's take a step back. How did the `BackupCoordinator` get the `pluginProviderPackageName`? It injects the `StoragePlugin` in the constructor which provides this package name. Can't the `SettingsManager` do the same?

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
```diff
@@ -132,6 +144,15 @@ class SettingsManager(private val context: Context) {
 
     fun isBackupEnabled(packageName: String) = !blacklistedApps.contains(packageName)
 
+    fun isAppAllowedForBackup(packageName: String): Boolean {
+        // Check that the app is not blacklisted by the user
+        val enabled = isBackupEnabled(packageName)
+        if (!enabled) Log.w(TAG, "Backup of $packageName disabled by user.")
+        // We need to exclude the DocumentsProvider used to store backup data.
+        // Otherwise, it gets killed when we back it up, terminating our backup.
+        return enabled && packageName != pluginProviderPackageName
```
grote, line 153: Nitpick: If you add `&& packageName != pluginProviderPackageName` part above, then the `pluginProviderPackageName` gets logged as disabled as well (would need to change log message). Maybe doesn't need to be logged as warning as it is expected behavior?

app/src/main/java/com/stevesoltys/seedvault/transport/ConfigurableBackupTransport.kt
```diff
@@ -117,9 +122,9 @@ class ConfigurableBackupTransport internal constructor(private val context: Cont
 
     override fun isAppEligibleForBackup(
         targetPackage: PackageInfo,
-        isFullBackup: Boolean,
+        @Suppress("UNUSED_PARAMETER") isFullBackup: Boolean,
     ): Boolean {
-        return backupCoordinator.isAppEligibleForBackup(targetPackage, isFullBackup)
+        return settingsManager.isAppAllowedForBackup(targetPackage.packageName)
```
grote, line 127: It was actually intentional that everything from here passes through `backupCoordinator`. Could still leave it like this and just hand over to settingsManager there, then you don't need to inject it here.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -28,13 +28,13 @@ private const val LOG_MAX_PACKAGES = 100
  */
 internal class PackageService(
     private val context: Context,
-    private val backupManager: IBackupManager,
+    private val settingsManager: SettingsManager,
 ) {
 
     private val packageManager: PackageManager = context.packageManager
     private val myUserId = UserHandle.myUserId()
```
grote, line 35: Looks like this isn't needed anymore, right?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -28,13 +28,13 @@ private const val LOG_MAX_PACKAGES = 100
...
 
-    val eligiblePackages: Array<String>
+    val requestedPackages: Array<String>
```
grote, line 37: Would be nice to add a Kotlin do to this field as the name isn't immediately clear to me. 

It seems to be only used when manually requesting a backup, so I wonder if backups by the system are actually working as D2D now as well. Did you try this?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -159,7 +164,13 @@ internal fun PackageInfo.isSystemApp(): Boolean {
...
 internal fun PackageInfo.allowsBackup(): Boolean {
     if (packageName == MAGIC_PACKAGE_MANAGER || applicationInfo == null) return false
-    return applicationInfo.flags and FLAG_ALLOW_BACKUP != 0
+
+    // At backup time, the system will filter out any apps that *it* does not want to be backed up.
+    // Now that we have switched to D2D, *we* generally want to back up as much as possible;
+    // part of the point of D2D is to ignore FLAG_ALLOW_BACKUP (allowsBackup). So, we return true.
+    // See frameworks/base/services/backup/java/com/android/server/backup/utils/
+    // BackupEligibilityRules.java lines 74-81 and 163-167 (tag: android-13.0.0_r8).
+    return true
```
grote, line 173: If we always return `true` here, I wonder, how that influences the other methods in this class that make use of it.

For example, will we still ever show "Apps that do not allow data backup" in the UI? If not, can we remove the related UI code and strings already?

Do we need to update the comment in other methods that still refer to backup not being allowed?

If we *almost* always return `true` here, could we maybe just eliminate this method completely? Maybe even other methods that are calling it, or at least simplifying them?

Sorry for being a pain in the ass here, but I think it is important that when implementing a new feature, it goes all the way. Just hacking the important bits in quickly will amass technical debt for future developers to solve.

grote (changes requested):

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
t-m-w: OK, stepping back... One thing at a time! :slightly_smiling_face: 

`BackupCoordinator` is not mocked in `BackupCoordinatorTest`; it is provided with the constructor arguments it needs through its actual constructor. `SettingsManager`, on the other hand, is mocked in tests. I'm not sure if it can receive constructor arguments while mocked -- I've been reading the mockk documentation without luck so far, but I may be missing something.

So, why not construct a real `SettingsManager`? From my understanding, unlike `BackupCoordinator`, `SettingsManager` attempts to access Android-exclusive things during initialization; if not mocked, it would require instrumented tests. And if we shift the code to provide  `plugin` via constructor, it will be `null` in testing because it is never initialized. If it is provided through the constructor as a `val`, we cannot assign to it -- but might be able to do something like `@VisibleForTesting(otherwise = PRIVATE)` on the setter. I'm not sure if that is acceptable.

grote: > SettingsManager, on the other hand, is mocked in tests. I'm not sure if it can receive constructor arguments while mocked

You may be misunderstanding the purpose of mocks in tests. We mock a dependency, so we don't need to worry about the actual implementation, its constructor etc., but can just focus on the interaction with the class.

> So, why not construct a real SettingsManager?

Because you don't want a full integration test here, but just unit test the `BackupCoordinator`. 

> And if we shift the code to provide plugin via constructor, it will be null in testing because it is never initialized.

Why? For a mocked `SettingsManager`, its internals do not matter at all, we just work on the API if you wish, method signatures, etc.

For the real `SettingsManager`, the plugin would be provided by the injection framework, the same way as it is now for `BackupCoordinator`.

grote: > If it is provided through the constructor as a val, we cannot assign to it -- but might be able to do something like @VisibleForTesting(otherwise = PRIVATE) on the setter. I'm not sure if that is acceptable.

No, oh god, no, please no! This is the same kind of hack as above. The `val`/`final` in the constructor is intentional, so you can be sure (compile-time sure) it will always be there and you don't rely on external components to set it.

t-m-w: Unfortunately, Seedvault blows up *today* from a cyclic dependency when trying to pass `StoragePlugin` to `SettingsManager` and running Seedvault on-device.

 `DocumentsProviderStoragePlugin` (which is the type of `StoragePlugin` we want to pass to `SettingsManager`) already depends on our `SettingsManager` itself. Its constructor takes a single `DocumentsStorage` object, and *that* object's constructor takes our single `SettingsManager`. If we try to inject the `StoragePlugin` in App.kt using `get()`, the app hangs until it crashes with a stack overflow.

~~So it looks like some significant refactoring will need to happen to accomplish the desired change.~~

Maybe just putting the method in `PackageService` and having `BackupCoordinator` call out to that is the most sensible, and it seems to work. I'll see how the other (manual) tests we've discussed go.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
nettnikl: Hi, since this is the 'do it properly' branch, i'd like to give my input here: i'd like to include all apps, but maybe, not everyone does. Especially since i know that app developers (eg of davx5) set allowBackup to false, as they did now want to give the false impression their app would work perfectly after a backup. I'd favor (in a polished, perfect solution) an option to say something like 'force include all apps'.

stevesoltys:
Resolved via #562.

Issue #476, opened 2022-11-21
Turning off "Backup my apps" prevents existing backup from being restored


t-m-w:
When turning off "Backup my apps", the existing backup set becomes unusable for restore. Attempting to restore the set follows through the usual process of entering the recovery key, but then leads to this message: "No suitable backups found at given location" and the following in logcat:
```
11-16 19:07:19.056 10138 29491 29709 I RestoreCoordinator: Got available metadata for tokens: [1668641603722]
11-16 19:07:19.056 10138 29491 29709 D RestoreViewModel: Ignoring RestoreSet with no last backup time: 1668641603722.
```

I can replicate this both with USB backup/restore and Internal Storage backup/restore. It has happened when restoring to the main user and restoring to a work profile. It is reproducible whenever "Backup my apps" had been turned off.

## Debugging
After turning off "Backup my apps", Seedvault [disables the backup service](https://github.com/seedvault-app/seedvault/blob/271b9a560f6ce4ebf1e778727c348ed364497c57/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsFragment.kt#L67). For some reason, this (or something else related to turning this off) leads to BackupManagerService re-initializing the transport, which reinitializes the device, causing `BackupMetadata` to be cleared. The `time` field for the existing backup set becomes zero, and the above error occurs. (See logcat snippet further down.)

~~The documentation for [IBackupManager.setBackupEnabled()](https://cs.android.com/android/platform/superproject/+/android-13.0.0_r8:frameworks/base/core/java/android/app/backup/IBackupManager.aidl;l=142-161) doesn't suggest to me that this should happen; in fact, to me, it suggests the opposite, by discussing how data-changed notifications will continue. Nevertheless, it does happen.~~

This appears to be intended behavior on the AOSP side, when disabling backup. See [UserBackupManagerService.updateStateOnBackupEnabled()](https://cs.android.com/android/platform/superproject/+/android-13.0.0_r8:frameworks/base/services/backup/java/com/android/server/backup/UserBackupManagerService.java;l=3195-3197).

To work around this, perhaps Seedvault could disregard initialization calls from AOSP when backup is disabled, or - more simply - take some measures to ensure that the backup set is not modified when "Backup my apps" is turned off, unless fully intended.

<details open><summary>logcat when turning off "Backup my apps"</summary>

```
 1000  2201  2766 I BackupManagerService: [UserID:0] Backup enabled => false
[...]

PR #477, opened 2022-11-22
Preserve backups when disabling "Backup my apps"

t-m-w:
Ignore AOSP's attempt to wipe backup data when backups are disabled.

Remaining quirks:
* When backups are re-enabled and another backup is started, the system will call initializeDevice, and Seedvault will generate a new backup salt and start the backup from scratch. This was already the case, and this patch does not change that.

Issue: seedvault-app/seedvault#476
Change-Id: I6ab41a885fcf7c4143814ebe849b8263a4f6e595


chirayudesai:
Can you also do the scary warning dialog when turning off backups?

That we can merge separately (and safely, early) since it'd just be a UI change.

chirayudesai:
https://github.com/seedvault-app/seedvault/pull/496 is a simpler solution

PR #479, opened 2022-11-24
Update permissions for log feature

t-m-w:
CalyxOS fails to boot with the new log changes, without android.permission.READ_LOGS.


chirayudesai (approved):

grote:
Ooops. Why did that work on my test device?

Do we still get the permission dialog now?

chirayudesai:
> Ooops. Why did that work on my test device?

Because you adb installed. ROM builds need this

> Do we still get the permission dialog now?

Edit hold on, will double check with fresh install.

chirayudesai:
> Edit hold on, will double check with fresh install.

Multiple of us checked, you always get the permission dialog.

Issue #247, opened 2021-05-03
Permission denial after choosing backup folder


skrimix:
I'm trying to install Seedvault as standalone app on AOSP-based custom rom, for this I made magisk module. The app and two xml files install fine, but after I choose backup folder in DocumentsUI an error message is shown: "Unable to get the permission to write to the backup location."
Logcat shows these errors:
```
05-03 16:51:23.308 20168 20168 E com.stevesoltys.seedvault.ui.storage.StorageActivity: OpenDocumentTree: content://com.android.externalstorage.documents/tree/primary%3ASeedVault
05-03 16:51:23.320 19526 19556 E DatabaseUtils: Writing exception to parcel
05-03 16:51:23.320 19526 19556 E DatabaseUtils: java.lang.SecurityException: Permission Denial: reading com.android.externalstorage.ExternalStorageProvider uri content://com.android.externalstorage.documents/root from pid=20168, uid=10162 requires that you obtain access using ACTION_OPEN_DOCUMENT or related APIs
05-03 16:51:23.320 19526 19556 E DatabaseUtils: 	at android.content.ContentProvider.enforceReadPermissionInner(ContentProvider.java:820)
05-03 16:51:23.320 19526 19556 E DatabaseUtils: 	at com.android.externalstorage.ExternalStorageProvider.enforceReadPermissionInner(ExternalStorageProvider.java:147)
05-03 16:51:23.320 19526 19556 E DatabaseUtils: 	at android.content.ContentProvider$Transport.enforceReadPermission(ContentProvider.java:684)
05-03 16:51:23.320 19526 19556 E DatabaseUtils: 	at android.content.ContentProvider$Transport.query(ContentProvider.java:239)
05-03 16:51:23.320 19526 19556 E DatabaseUtils: 	at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:106)
05-03 16:51:23.320 19526 19556 E DatabaseUtils: 	at android.os.Binder.execTransactInternal(Binder.java:1154)
05-03 16:51:23.320 19526 19556 E DatabaseUtils: 	at android.os.Binder.execTransact(Binder.java:1123)
05-03 16:51:23.321 20168 20168 W StorageRootResolver: Failed to load some roots from com.android.externalstorage.documents
[...]

PR #480, opened 2022-11-28
[UNTESTED] Support a lack of MANAGE_DOCUMENTS permission

t-m-w:
The README.md and code comments indicate that it should be possible to use Seedvault without `android.permission.MANAGE_DOCUMENTS`, but that may not be accurate, currently.

When `MANAGE_DOCUMENTS` is not granted, Seedvault prompts to select a storage location using the documents navigator. No matter what location is chosen, Seedvault cannot find backups at that location and experiences errors in logcat as described in seedvault-app/seedvault#247.

With this patch, if resolving the storage's actual root fails, a "fake" storage root is used, as is already done for Nextcloud, Davx5, and USB. (It might be the case that there is no use even *trying* to resolve the storage's actual root, for this situation.)

To test:
1. Include a Seedvault apk that has not been signed with the platform signing key so that it will not be granted the `MANAGE_DOCUMENTS` permission. (One scenario that would cause this would be attempting to include Seedvault in an otherwise-unsupported ROM via a Magisk module.)
2. Try to restore a backup set via `*#*#RESTORE#*#*`.
3. Browse to the parent location of a .SeedVaultAndroidBackup folder.
4. Enter the recovery key when prompted.
5. ...see if it works beyond this point.

(Until now, I have only been able to test up to the point that I am prompted for a recovery key.)

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageActivity.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageActivity.kt
@@ -35,7 +35,9 @@ class StorageActivity : BackupActivity() {
             val authority = uri.authority ?: throw AssertionError("No authority in $uri")
             val storageRoot = StorageRootResolver.getStorageRoots(this, authority).getOrNull(0)
             if (storageRoot == null) {
-                viewModel.onUriPermissionResultReceived(null)
+                viewModel.onSafOptionChosen(
+                    StorageRootResolver.getFakeStorageRootForUri(this, uri))
+                viewModel.onUriPermissionResultReceived(uri)
             } else {
                 viewModel.onSafOptionChosen(storageRoot)
                 viewModel.onUriPermissionResultReceived(uri)
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
@@ -3,6 +3,7 @@ package com.stevesoltys.seedvault.ui.storage
 import android.content.Context
 import android.database.Cursor
 import android.graphics.drawable.Drawable
+import android.net.Uri
 import android.os.UserHandle
 import android.provider.DocumentsContract
 import android.provider.DocumentsContract.Root.COLUMN_AVAILABLE_BYTES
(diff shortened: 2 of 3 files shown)


chirayudesai:
1. We could setup CI to build and release APKs without this perm - maybe have a flavor?
2. Users have always asked to be able to choose a folder, especially those who sync their backups.

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
```diff
@@ -57,6 +58,21 @@ internal object StorageRootResolver {
         return roots
     }
 
+    fun getFakeStorageRootForUri(context: Context, uri: Uri): SafOption {
+        return SafOption(
+            authority = AUTHORITY_STORAGE,
+            rootId = "fake",
+            documentId = "fake",
+            // TODO: Use something other than the USB icon?
+            icon = getIcon(context, AUTHORITY_STORAGE, "usb", 0),
```
grote, line 67: Yeah a different icon would be better. But is this even shown to the user somewhere?

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageActivity.kt
```diff
@@ -35,7 +35,9 @@ class StorageActivity : BackupActivity() {
             val authority = uri.authority ?: throw AssertionError("No authority in $uri")
             val storageRoot = StorageRootResolver.getStorageRoots(this, authority).getOrNull(0)
             if (storageRoot == null) {
-                viewModel.onUriPermissionResultReceived(null)
+                viewModel.onSafOptionChosen(
+                    StorageRootResolver.getFakeStorageRootForUri(this, uri))
+                viewModel.onUriPermissionResultReceived(uri)
```
grote, line 40: Oops, I guess that must have broken in some recent refactoring.

Btw. I don't really like this code in the activity. Maybe we should move it into the ViewModel and start (unit-) testing critical code paths like this.

grote:
What's the status here? @t-m-w do you want to do another pass or should I pick this up?

t-m-w:
Copy of last comment :slightly_smiling_face:

Either way works for me. I likely won't get a chance to look at Seedvault stuff until Tuesday of next week at the earliest.

grote:
closed in favor of https://github.com/seedvault-app/seedvault/pull/551

PR #484, opened 2022-12-05
[Needs More Testing] Allow restoring data for a greater number of apps

t-m-w:
Restore app data in smaller batches when performing a full restoration from a backup set, to prevent a Binder exception that causes the process to fail entirely.

Android may encounter this exception when trying to call the transport.startRestore() method if too many packages are involved; in testing, 300 is an example of too many. Instead of using IRestoreSession.restoreAll(), use restorePackages() and provide the package names in batches of 100.

This issue reveals itself when using SeedVault with the D2D patch and with an OS, such as stock Pixel OS, that includes an abundance of packages. (Prior to this patch, the call to restoreAll() meant that the framework would request data restoration for all packages installed, even if they were not in the metadata.)

In logs, this issue appears as follows:
```
I BackupManagerService: Full restore; asking about 300 apps
W BpBinder: Large outgoing transaction of 528540 bytes, interface descriptor <uncached descriptor>, code 14
E JavaBinder: !!! FAILED BINDER TRANSACTION !!!  (parcel size = 528540)
E BackupManagerService: Unable to contact transport for restore: data parcel size 528540 bytes
```

Change-Id: Ibb5bb4572d9e873beccd6056da5fe3ae4dce71c2

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
@@ -64,8 +64,11 @@ import org.calyxos.backup.storage.restore.RestoreService.Companion.EXTRA_TIMESTA
 import org.calyxos.backup.storage.restore.RestoreService.Companion.EXTRA_USER_ID
 import org.calyxos.backup.storage.ui.restore.SnapshotViewModel
 import java.util.LinkedList
+import kotlin.math.min
 
 private val TAG = RestoreViewModel::class.java.simpleName
+
+internal const val PACKAGES_PER_CHUNK = 100
 
 internal class RestoreViewModel(
     app: Application,
@@ -201,18 +204,11 @@ internal class RestoreViewModel(
         }
 
         // we need to retrieve the restore sets before starting the restore
-        // otherwise restoreAll() won't work as they need the restore sets cached internally
-        val observer = RestoreObserver { observer ->
-            // this lambda gets executed after we got the restore sets
-            // now we can start the restore of all available packages
-            val restoreAllResult = session.restoreAll(token, observer, monitor)
-            if (restoreAllResult != 0) {
-                Log.e(TAG, "restoreAll() returned non-zero value")
-                mRestoreBackupResult.postValue(
-                    RestoreBackupResult(app.getString(R.string.restore_finished_error))
-                )
-            }
-        }
+        // otherwise restorePackages() won't work as they need the restore sets cached internally
+        val packages = mChosenRestorableBackup.value?.packageMetadataMap?.keys?.toMutableList()
+            ?: mutableListOf()
+        // The chunked restoration is performed within the RestoreObserver.
+        val observer = RestoreObserver(session, token, packages, monitor)
         if (session.getAvailableRestoreSets(observer, monitor) != 0) {
             Log.e(TAG, "getAvailableRestoreSets() returned non-zero value")
             mRestoreBackupResult.postValue(
@@ -294,8 +290,12 @@ internal class RestoreViewModel(
     }
 
     @WorkerThread
-    private inner class RestoreObserver(private val startRestore: (RestoreObserver) -> Unit) :
-        IRestoreObserver.Stub() {
+    private inner class RestoreObserver(
+        private val session: IRestoreSession,
+        private val token: Long,
+        private val packages: MutableList<String>,
+        private val monitor: BackupMonitor
+    ) : IRestoreObserver.Stub() {
 
         /**
          * Supply a list of the restore datasets available from the current transport.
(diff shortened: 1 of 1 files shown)


chirayudesai (approved):

grote:
Wow nice find! But WTF? You'd imagine the Android backup API being able work and not die with lots of apps when using a perfectly reasonable API call.

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
```diff
@@ -201,18 +204,11 @@ internal class RestoreViewModel(
         }
 
         // we need to retrieve the restore sets before starting the restore
-        // otherwise restoreAll() won't work as they need the restore sets cached internally
-        val observer = RestoreObserver { observer ->
```
grote, line 205: Nice, so we don't need to retrieve the restore sets before anymore?

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
```diff
@@ -294,8 +290,12 @@ internal class RestoreViewModel(
     }
 
     @WorkerThread
-    private inner class RestoreObserver(private val startRestore: (RestoreObserver) -> Unit) :
-        IRestoreObserver.Stub() {
+    private inner class RestoreObserver(
+        private val session: IRestoreSession,
+        private val token: Long,
+        private val packages: MutableList<String>,
```
grote, line 296: Does the list need to be mutable here? (Good practice to build API where the recipient can't modify your objects if they don't need to)

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
```diff
@@ -307,7 +307,22 @@ internal class RestoreViewModel(
...
+            // this gets executed after we got the restore sets
+            // now we can start the restore of all available packages
+            restoreNextPackages()
+        }
+
+        private fun restoreNextPackages() {
+            // Packages are restored in chunks, or else transport.startRestore() in the framework's
+            // PerformUnifiedRestoreTask.java may fail due to an oversized Binder transaction,
+            // causing the entire restoration to fail.
+            val packagesChunk = packages.subList(0, min(packages.size, PACKAGES_PER_CHUNK))
```
grote, line 319: Isn't this always returning the same number of packages running into an infinite loop or what am I missin ghere?

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
```diff
@@ -307,7 +307,22 @@ internal class RestoreViewModel(
...
+            val result =
+                session.restorePackages(token, this, packagesChunk.toTypedArray(), monitor)
+            packagesChunk.clear()
```
grote, line 322: why do you need to clear the chunks?

grote: Is this some implicit modification of the `packages` list? If so, I don't like it and would prefer something more easy to read and understand.

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
```diff
@@ -64,9 +64,12 @@ import org.calyxos.backup.storage.restore.RestoreService.Companion.EXTRA_TIMESTA
 import org.calyxos.backup.storage.restore.RestoreService.Companion.EXTRA_USER_ID
 import org.calyxos.backup.storage.ui.restore.SnapshotViewModel
 import java.util.LinkedList
+import kotlin.math.min
 
 private val TAG = RestoreViewModel::class.java.simpleName
 
+internal const val PACKAGES_PER_CHUNK = 100
```
grote, line 71: I guess the challenge here is finding the right number. I've seen the binder transaction limit vary greatly depending on what's going on otherwise with the system.

grote (changes requested):

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
chirayudesai: We can try to pick something conservative here and just try to tweak it if needed later.

* Now: Lots of apps may fail
* With this: More likely to succeed
* In the future, if needed: decrease this further?

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
t-m-w: Am I missing something, or are you saying this because the comment appears to have been removed? It wasn't removed, it was modified to say `restorePackages()` instead of `restoreAll()`, since this changes what we are calling.

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
t-m-w: It's not like we're going to get any tangible benefit from setting it very high, right? Maybe we could just set it to 25 or something?

app/src/main/java/com/stevesoltys/seedvault/restore/RestoreViewModel.kt
grote: Ah it is down there. So we still need to do that, did you try?

grote:
What's the status here? @t-m-w do you want to do another pass or should I pick this up?

t-m-w:
Either way works for me. I likely won't get a chance to look at Seedvault stuff until Tuesday of next week at the earliest.

chirayudesai:
All yours @grote

grote:
closing in favor of https://github.com/seedvault-app/seedvault/pull/563

PR #493, opened 2023-01-03
Warn before turning off backups

t-m-w:
Edit: Modified based on #496.

Let the user know that turning off backups will require a new backup set to be created.

Requires: #496 
Issue: seedvault-app/seedvault#476

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -14,6 +14,9 @@
     <string name="settings_backup_location_none">None</string>
     <string name="settings_backup_location_internal">Internal storage</string>
     <string name="settings_backup_last_backup_never">Never</string>
+    <string name="settings_backup_dialog_title">Really turn off app backups?</string>
+    <string name="settings_backup_dialog_message">When you turn backups on again, the backup process may take longer than usual, and it will use additional storage space.</string>
+    <string name="settings_backup_dialog_disable">Turn off anyway</string>
     <string name="settings_info">All backups are encrypted on your phone. To restore from backup you will need your 12-word recovery code.</string>
     <string name="settings_auto_restore_title">Automatic restore</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -14,6 +14,9 @@
     <string name="settings_backup_location_none">None</string>
     <string name="settings_backup_location_internal">Internal storage</string>
     <string name="settings_backup_last_backup_never">Never</string>
+    <string name="settings_backup_dialog_title">Really turn off backup?</string>
+    <string name="settings_backup_dialog_message">Turning off backup will... [prevent any currently-accessible backups from being restorable in the future] (fill this out later)</string>
```
chirayudesai: So it's either this, or once #477 is merged, we can say something about a brand new backup being created when re-enabled / mention the new restore set.

Ideally we don't block the dialog on the other change for too long.

t-m-w:
Updated strings in accordance with new PR. Do they sound reasonable?

app/src/main/res/values/strings.xml
```diff
@@ -14,6 +14,9 @@
     <string name="settings_backup_location_none">None</string>
     <string name="settings_backup_location_internal">Internal storage</string>
     <string name="settings_backup_last_backup_never">Never</string>
+    <string name="settings_backup_dialog_title">Really turn off app backups?</string>
+    <string name="settings_backup_dialog_message">When you turn backups on again, the backup process may take longer than usual, and it will use additional storage space.</string>
```
chirayudesai, line 18: This is technically correct. I think it sounds ok too. We can change this in the future when we add restore set management, since then we'll have terminology for that and a way for the user to actually do something about it.

chirayudesai (commented):
Untested, LGTM, should be shipped alongwith #496

chirayudesai (approved):
Good to go, just waiting on the dependent change so that the strings are valid.

Could also just merge it now to get some translations ready.

PR #495, opened 2023-01-03
Pre-grant ACCESS_MEDIA_LOCATION permission

grote:


chirayudesai (approved):
`fixed="true"` here should mean that it can't be revoked.

TBH we could also do the same for notifications in that case.

PR #496, opened 2023-01-03
Always start a new RestoreSet when initializing the device

grote:
This avoids deleting the current backup when the user disables backups (or the system decides to do a random re-init).

It is a maybe less hacky alternative to #477.

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
@@ -88,13 +88,16 @@ internal class BackupCoordinator(
      * Starts a new [RestoreSet] with a new token (the current unix epoch in milliseconds).
      * Call this at least once before calling [initializeDevice]
      * which must be called after this method to properly initialize the backup transport.
+     *
+     * @return the token of the new [RestoreSet].
      */
     @Throws(IOException::class)
-    suspend fun startNewRestoreSet() {
+    private suspend fun startNewRestoreSet(): Long {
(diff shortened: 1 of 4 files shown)


chirayudesai (commented):
Untested but LGTM.

As mentioned elsewhere, this seems far better than the alternative where it would destroy perfectly fine backups.

The worst case now is it creates more restore sets, and that's confusing when restoring, but not that much given it also displays the size and time of backup.

Finding out a backup was destroyed when trying to restore would be far worse (which this should hopefully avoid)

chirayudesai:
Test:

0. Create backup to internal storage without this change.
1. Install app with this.
2. Turn off backup, turn it back on.
3. Watch new restore set get created. (well, empty dir)
4. Menu -> Backup Now - another set gets created, dir gets filled.
5. Repeat 2-4, watch 2 sets (1 empty dir, 1 full) get created each time.

chirayudesai:
I don't mind merging this even with the above given that it's just empty dirs and won't show up when restoring.

chirayudesai (approved):

grote:
> Watch new restore set get created. (well, empty dir)
> Menu -> Backup Now - another set gets created, dir gets filled.

We should probably look into why yet another dir gets created here. Could do it in this PR or in a new one.

chirayudesai:
> > Watch new restore set get created. (well, empty dir)
> > Menu -> Backup Now - another set gets created, dir gets filled.
> 
> We should probably look into why yet another dir gets created here. Could do it in this PR or in a new one.

Let's do another PR, given this solves a critical issue. An extra empty dir is not that big a deal.

Issue #78, opened 2020-03-10
Seedvault bypasses device policy for installing apps


thestinger:
There are 3 relevant profile restrictions: [`DISALLOW_INSTALL_APPS`](https://developer.android.com/reference/android/os/UserManager#DISALLOW_INSTALL_APPS), [`DISALLOW_INSTALL_UNKNOWN_SOURCES`](https://developer.android.com/reference/android/os/UserManager#DISALLOW_INSTALL_UNKNOWN_SOURCES) and [`DISALLOW_INSTALL_UNKNOWN_SOURCES_GLOBALLY`](https://developer.android.com/reference/android/os/UserManager#DISALLOW_INSTALL_UNKNOWN_SOURCES_GLOBALLY).

Seedvault currently provides a bypass for these restrictions, by using the privileged permission for installing apps without enforcing these. These are core features for locking down profiles as part of device management, either for an enterprise deployment or by the owner of the device. Any app from the backup needs to be considered as coming from an unknown source, so if any of these user profile restrictions is active for the profile Seedvault should disable installing apps from the backup.

chirayudesai:
We'd also need to account for `DISALLOW_UNINSTALL_APPS` for the auto-restore notification which has an "uninstall" button.

grote:
Is there another install API we could use that already respects these profile restrictions (that doesn't require user-interaction) or there just this one API and it is it simply our responsibility to observe these restrictions when we use it?

Normally, Seedvault restore and app-install capabilities are only meant to be used during setupwizard. Does this change the picture maybe? Are restrictions in effect before the device is provisioned?

> We'd also need to account for DISALLOW_UNINSTALL_APPS for the auto-restore notification which has an "uninstall" button.

This button however makes use of a deprecated API that requires user-interaction for uninstall, so maybe that already observes the restriction?

thestinger:
> Normally, Seedvault restore and app-install capabilities are only meant to be used during setupwizard. Does this change the picture maybe? Are restrictions in effect before the device is provisioned?

I don't think so, since the best way of setting up a device manager is doing it as the first step of provisioning to make sure the restrictions are enforced from that point on. That's how Google's own enterprise device management works.

thestinger:
> Is there another install API we could use that already respects these profile restrictions (that doesn't require user-interaction) or there just this one API and it is it simply our responsibility to observe these restrictions when we use it?

[...]

PR #502, opened 2023-01-12
Skip installing APKs if not allowed by policy

chirayudesai:
Issue: #78 

When APK install is blocked by policy, this should treat it the same as if the APK just wasn't present, and then use the flow we already have for that ("Tap to install"), which is normally used when people turn off "APK Backup"

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
@@ -12,6 +12,7 @@ import com.stevesoltys.seedvault.plugins.LegacyStoragePlugin
 import com.stevesoltys.seedvault.plugins.StoragePlugin
 import com.stevesoltys.seedvault.plugins.StoragePluginManager
 import com.stevesoltys.seedvault.restore.RestorableBackup
+import com.stevesoltys.seedvault.restore.install.ApkInstallState.FAILED
 import com.stevesoltys.seedvault.restore.install.ApkInstallState.FAILED_SYSTEM_APP
 import com.stevesoltys.seedvault.restore.install.ApkInstallState.IN_PROGRESS
 import com.stevesoltys.seedvault.restore.install.ApkInstallState.QUEUED
@@ -35,6 +36,7 @@ internal class ApkRestore(
     private val crypto: Crypto,
     private val splitCompatChecker: ApkSplitCompatibilityChecker,
     private val apkInstaller: ApkInstaller,
+    private val installRestriction: InstallRestriction,
 ) {
 
     private val pm = context.packageManager
@@ -47,6 +49,7 @@ internal class ApkRestore(
             // Otherwise, it gets killed when we install it, terminating our restoration.
             it.key != storagePlugin.providerPackageName
         }
+        val isAllowedToInstallApks = installRestriction.isAllowedToInstallApks()
         val total = packages.size
         var progress = 0
 
@@ -57,11 +60,17 @@ internal class ApkRestore(
             installResult[packageName] = ApkInstallResult(
                 packageName = packageName,
                 progress = progress,
-                state = QUEUED,
+                state = if (isAllowedToInstallApks) QUEUED else FAILED,
                 installerPackageName = metadata.installer
             )
         }
-        emit(installResult)
+        if (isAllowedToInstallApks) {
+            emit(installResult)
+        } else {
+            installResult.isFinished = true
+            emit(installResult)
+            return@flow
+        }
 
         // re-install individual packages and emit updates
         for ((packageName, metadata) in packages) {
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/install/InstallModule.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/install/InstallModule.kt
@@ -1,5 +1,6 @@
 package com.stevesoltys.seedvault.restore.install
 
+import android.os.UserManager
 import org.koin.android.ext.koin.androidContext
 import org.koin.dsl.module
 
@@ -7,5 +8,9 @@ val installModule = module {
     factory { ApkInstaller(androidContext()) }
     factory { DeviceInfo(androidContext()) }
     factory { ApkSplitCompatibilityChecker(get()) }
-    factory { ApkRestore(androidContext(), get(), get(), get(), get(), get()) }
+    factory {
+        ApkRestore(androidContext(), get(), get(), get(), get(), get()) {
+            androidContext().getSystemService(UserManager::class.java).isAllowedToInstallApks()
+        }
+    }
 }
--- a/app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkBackupRestoreTest.kt
+++ b/app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkBackupRestoreTest.kt
@@ -52,6 +52,7 @@ internal class ApkBackupRestoreTest : TransportTest() {
     private val storagePlugin: StoragePlugin<*> = mockk()
     private val splitCompatChecker: ApkSplitCompatibilityChecker = mockk()
     private val apkInstaller: ApkInstaller = mockk()
+    private val installRestriction: InstallRestriction = mockk()
 
     private val apkBackup = ApkBackup(pm, crypto, settingsManager, metadataManager)
     private val apkRestore: ApkRestore = ApkRestore(
@@ -60,7 +61,8 @@ internal class ApkBackupRestoreTest : TransportTest() {
         legacyStoragePlugin = legacyStoragePlugin,
         crypto = crypto,
         splitCompatChecker = splitCompatChecker,
-        apkInstaller = apkInstaller
+        apkInstaller = apkInstaller,
+        installRestriction = installRestriction,
     )
 
     private val signatureBytes = byteArrayOf(0x01, 0x02, 0x03)
@@ -132,6 +134,7 @@ internal class ApkBackupRestoreTest : TransportTest() {
         val apkPath = slot<String>()
         val cacheFiles = slot<List<File>>()
 
+        every { installRestriction.isAllowedToInstallApks() } returns true
         every { strictContext.cacheDir } returns tmpFile
         every { crypto.getNameForApk(salt, packageName, "") } returns name
         coEvery { storagePlugin.getInputStream(token, name) } returns inputStream
--- a/app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkRestoreTest.kt
+++ b/app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkRestoreTest.kt
@@ -54,6 +54,7 @@ internal class ApkRestoreTest : TransportTest() {
     private val legacyStoragePlugin: LegacyStoragePlugin = mockk()
     private val splitCompatChecker: ApkSplitCompatibilityChecker = mockk()
     private val apkInstaller: ApkInstaller = mockk()
+    private val installRestriction: InstallRestriction = mockk()
 
     private val apkRestore: ApkRestore = ApkRestore(
         context = strictContext,
@@ -62,6 +63,7 @@ internal class ApkRestoreTest : TransportTest() {
         crypto = crypto,
         splitCompatChecker = splitCompatChecker,
         apkInstaller = apkInstaller,
+        installRestriction = installRestriction,
     )
 
     private val icon: Drawable = mockk()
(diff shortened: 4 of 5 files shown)


app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
```diff
@@ -37,15 +38,22 @@ internal class ApkRestore(
 ) {
 
     private val pm = context.packageManager
+    private val restrictions = context.getSystemService(UserManager::class.java).userRestrictions
```
chirayudesai: Need to figure out how to mock this

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
```diff
@@ -37,15 +38,22 @@ internal class ApkRestore(
...
+            false) ||
+            restrictions.getBoolean(UserManager.DISALLOW_INSTALL_UNKNOWN_SOURCES_GLOBALLY,
+            false)
         // filter out packages without APK and get total
         val packages = backup.packageMetadataMap.filter {
             // We also need to exclude the DocumentsProvider used to retrieve backup data.
             // Otherwise, it gets killed when we install it, terminating our restoration.
             val isStorageProvider = it.key == storagePlugin.providerPackageName
-            it.value.hasApk() && !isStorageProvider
+            it.value.hasApk() && !isStorageProvider && !cantInstallApks
```
chirayudesai: I'm sure there's a more elegant way to do this.

Maybe check this, and then init an empty `MutableInstallResult`, set to finish and return that?

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
```diff
@@ -37,15 +38,22 @@ internal class ApkRestore(
...
+    private val restrictions = context.getSystemService(UserManager::class.java).userRestrictions
 
     @Suppress("BlockingMethodInNonBlockingContext")
     fun restore(backup: RestorableBackup) = flow {
+        val cantInstallApks = restrictions.getBoolean(UserManager.DISALLOW_INSTALL_APPS,
+            false) ||
+            restrictions.getBoolean(UserManager.DISALLOW_INSTALL_UNKNOWN_SOURCES,
+            false) ||
+            restrictions.getBoolean(UserManager.DISALLOW_INSTALL_UNKNOWN_SOURCES_GLOBALLY,
+            false)
```
chirayudesai: Need to double check if this is how we want to check for it - (false is defaultValue)

chirayudesai (commented):
Untested but I thought of just putting it up for a start and then it can be worked on by anyone.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
grote: Maybe pass it as a constructor parameter via injection, so you can mock it here.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
grote: That is a bit hard to read. Maybe define it as an extension function on `userRestrictions`, so you can mock its return value for testing?

grote: Could also statically import the `UserManager.*` constants.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
grote: To do what? If we can not install any APK, we don't need to do any filtering, right?

We could probably emit an install result right away with all apps as failed (if they are not already installed)?

chirayudesai: > To do what? If we can not install any APK, we don't need to do any filtering, right?

Yes

> We could probably emit an install result right away with all apps as failed (if they are not already installed)?

Yes that was the other option I considered.

app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkBackupRestoreTest.kt
```diff
@@ -89,6 +92,10 @@ internal class ApkBackupRestoreTest : TransportTest() {
         if (name == this.name) outputStream else splitOutputStream
     }
 
+    fun UserManager.checkRestriction(restriction: String): Boolean {
+        return this.userRestrictions.getBoolean(restriction, false)
```
chirayudesai: This needs to get changed.

app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkRestoreTest.kt
```diff
@@ -82,6 +85,10 @@ internal class ApkRestoreTest : TransportTest() {
     private val backup = RestorableBackup(metadata.copy(packageMetadataMap = packageMetadataMap))
     private val suffixName = getRandomString()
 
+    fun UserManager.checkRestriction(restriction: String): Boolean {
+        return this.userRestrictions.getBoolean(restriction, false)
```
chirayudesai: ditto.

grote: not very nice, but that's how to mock extension functions: https://mockk.io/#extension-functions

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
```diff
@@ -34,18 +38,29 @@ internal class ApkRestore(
...
+    private val userManager: UserManager,
 ) {
 
     private val pm = context.packageManager
 
+    fun UserManager.checkRestriction(restriction: String): Boolean {
+        return this.userRestrictions.getBoolean(restriction, false)
+    }
+
+    private fun cantInstallApks(): Boolean {
```
grote: maybe call it something like `isAllowedToInstallApks()`, because technically we can install, right?

with this the function it looks much more readable than before, btw.

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
```diff
@@ -34,18 +38,29 @@ internal class ApkRestore(
...
+        return userManager.checkRestriction(DISALLOW_INSTALL_APPS) ||
+            userManager.checkRestriction(DISALLOW_INSTALL_UNKNOWN_SOURCES) ||
+            userManager.checkRestriction(DISALLOW_INSTALL_UNKNOWN_SOURCES_GLOBALLY)
```
grote: Would Seedvault be considered an unknown source? Why?

chirayudesai: Yes. Because the backup can come from anywhere.

This was discussed in the issue further.

One could craft a backup with just one app, get SeedVault to restore it, and thus bypass this restriction entirely.

We don't have a way of telling legit backups apart, so we leave it upto the system by letting it decide which app it wants to allow to be installed (depending upon the store configured), and which app it doesn't.

app/src/test/java/com/stevesoltys/seedvault/restore/install/ApkRestoreTest.kt
```diff
@@ -403,6 +416,24 @@ internal class ApkRestoreTest : TransportTest() {
...
 
+    @Test
+    fun `no apks get installed when blocked by policy`(@TempDir tmpDir: Path) = runBlocking {
+        every { installRestriction.isAllowedToInstallApks() } returns false
+        every { storagePlugin.providerPackageName } returns storageProviderPackageName
+
+        apkRestore.restore(backup).collectIndexed { i, value ->
+            when (i) {
+                0 -> {
+                    // single package fails without attempting to install it
```
chirayudesai, line 430: @grote I didn't get this part, why the single package?

grote: because we are just testing with a single package here

chirayudesai: What I don't get is where that one package is coming from / what package would that be.

I do see that we expect it to fail installing, that makes sense.

grote: Follow the traces. We call `apkRestore.restore(backup)`.  The parameter is `val backup = RestorableBackup(metadata.copy(packageMetadataMap = packageMetadataMap))`. Then `packageMetadataMap = hashMapOf(packageName to packageMetadata)`, so only a single package being used for this test.

grote (dismissed):

t-m-w:
Downstream tracking issue: https://gitlab.com/CalyxOS/calyxos/-/issues/1519

chirayudesai:
Still needs testing :)

app/src/main/java/com/stevesoltys/seedvault/restore/install/InstallModule.kt
```diff
@@ -1,11 +1,16 @@
...
 import org.koin.dsl.module
 
 val installModule = module {
     factory { ApkInstaller(androidContext()) }
     factory { DeviceInfo(androidContext()) }
     factory { ApkSplitCompatibilityChecker(get()) }
-    factory { ApkRestore(androidContext(), get(), get(), get(), get(), get()) }
+    factory {
+        ApkRestore(androidContext(), get(), get(), get(), get(), get()) {
+            androidContext().getSystemService(UserManager::class.java).isAllowedToInstallApks()
```
t-m-w, line 13: When building in AOSP:
```
packages/apps/Seedvault/app/src/main/java/com/stevesoltys/seedvault/restore/install/InstallModule.kt:13:71: error: only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type UserManager?
            androidContext().getSystemService(UserManager::class.java).isAllowedToInstallApks()
                                                                      ^
```

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
```diff
@@ -55,11 +58,17 @@ internal class ApkRestore(
             installResult[packageName] = ApkInstallResult(
                 packageName = packageName,
                 progress = progress,
-                state = QUEUED,
+                state = if (isAllowedToInstallApks) QUEUED else FAILED,
```
t-m-w, line 63: So the intention is for this to appear to the user as a failure? Looks like it, and that is what is happening, including for system apps, which is fairly confusing.

<details><summary>Screenshots</summary>

Screenshot_20240416-063633

Screenshot_20240416-063652

app/src/main/java/com/stevesoltys/seedvault/restore/install/InstallModule.kt
grote: Could use `androidContext().getSystemService(UserManager::class.java)?.isAllowedToInstallApks() ?: false`

app/src/main/java/com/stevesoltys/seedvault/restore/install/ApkRestore.kt
grote: This may be the same if apk backup is disabled. I think we show failed there as well whenever we don't have an APK.

chirayudesai: Yep, this patch tries to just make the policy block same as "we don't have an APK"

Improvements to that definitely welcome separately. We haven't received much feedback about it in general, making me think most users choose to backup APKs.

t-m-w (dismissed):
Given what we discussed, and the fact that failing to install is intended behavior currently, I think this change is fine.

chirayudesai:
Rebased, one trivial conflict in tests fixed.

chirayudesai:
One test failure:

> > Task :app:connectedDebugAndroidTest
> Starting 1 tests on test(AVD) - 14
> 
> com.stevesoltys.seedvault.e2e.impl.BackupRestoreTest > backup and restore applications[test(AVD) - 14] FAILED 
> 	android.content.res.Resources$NotFoundException: Resource ID #0x7f060258 type #0x2 is not valid
> 	at android.content.res.Resources.getColor(Resources.java:1076)
> Tests on test(AVD) - 14 failed: There was 1 failure(s).

grote:
that test failure happens everywhere, we don't know why.

grote (approved):

PR #523, opened 2023-03-09
Icon updates

mikeNG:


app/src/main/res/values/colors.xml
```diff
@@ -8,6 +8,7 @@
     <color name="statusBarColor">@*android:color/primary_device_default_settings_light
     </color>
     <color name="red">@*android:color/error_color_device_default_dark</color>
+    <color name="ic_launcher_background">@*android:color/accent_device_default_light</color>
```
chirayudesai, line 11: Do we want to do this?

chirayudesai (changes requested):
Screenshot_20230310-001550

app/src/main/res/values/colors.xml
theimpulson: @grote I feel you might be better suited to make the decision here.

chirayudesai:
@mikeNG didn't we have some other issue with this apart from just the color (which is a quick one liner)

mikeNG:
The logo size is different than the original icon, that needs to be fixed

app/src/main/res/values/colors.xml
grote: I think this way of supplying the resources doesn't work with gradle builds, right?

theimpulson: Works fine on both Gradle and AOSP builds

theimpulson (approved):
Fine by me but I will leave final approval to @grote as the icon color is changed as per the system's accent color and that feels like a major change someone else should approve (in a UI design sense).

theimpulson:
> The logo size is different than the original icon, that needs to be fixed

It's slightly larger compared to the old one but still is fine and follows the adaptive icon guidelines. I will say it's an improvement.

chirayudesai (approved):

chirayudesai:
(I rebased this to get the new fancy CI to run)

chirayudesai:
Screengrab from emulator (Via CI)

Screenshot_2023-09-26_02-55-16

PR #526, opened 2023-03-27
Android13 QPR2 Dependency & Misc Updates

theimpulson:
This PR updates dependencies to the latest release (as per AOSP and Gradle) wherever possible.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ plugins {
     id 'org.jetbrains.kotlin.android' version "$kotlin_version" apply false
     id 'org.jetbrains.kotlin.kapt' version "$kotlin_version" apply false
     id 'org.jetbrains.dokka' version "$kotlin_version" apply false
-    id 'org.jlleitschuh.gradle.ktlint' version '10.2.0' apply false
+    id 'org.jlleitschuh.gradle.ktlint' version '11.5.0' apply false
 }
 
 ext {
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,6 @@ plugins {
 }
 
 ext {
-    buildToolsVersion = '33.0.0'
     compileSdkVersion = 33
     minSdkVersion = 32
     targetSdkVersion = 33
--- a/build.gradle
+++ b/build.gradle
@@ -20,9 +20,9 @@ plugins {
 }
 
 ext {
-    compileSdkVersion = 33
-    minSdkVersion = 32
-    targetSdkVersion = 33
+    compileSdk = 33
+    minSdk = 32
+    targetSdk = 33
 }
 
 apply from: 'gradle/dependencies.gradle'
--- a/build.gradle
+++ b/build.gradle
@@ -2,10 +2,11 @@ buildscript {
     // 1.3.61 Android 11
     // 1.4.30 Android 12
     // 1.6.10 Android 13
+    // 1.7.20 Android 13 (QPR2)
     // Check:
-    // https://android.googlesource.com/platform/external/kotlinc/+/refs/tags/android-13.0.0_r3/build.txt
-    ext.aosp_kotlin_version = '1.6.10'  // 1.6.10-release-923 in AOSP
-    ext.kotlin_version = '1.6.10'
+    // https://android.googlesource.com/platform/external/kotlinc/+/refs/tags/android-13.0.0_r32/build.txt
+    ext.aosp_kotlin_version = '1.7.20'  // 1.7.20-release-201 in AOSP
+    ext.kotlin_version = '1.7.20'
 
     repositories {
         mavenCentral()
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ buildscript {
         //noinspection DifferentKotlinGradleVersion
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "com.google.protobuf:protobuf-gradle-plugin:0.8.17"
-        classpath 'com.android.tools.build:gradle:7.2.2'
+        classpath 'com.android.tools.build:gradle:7.4.2'
     }
 }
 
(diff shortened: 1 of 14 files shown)


build.gradle
```diff
@@ -1,11 +1,11 @@
 buildscript {
     // 1.3.61 Android 11
     // 1.4.30 Android 12
-    // 1.6.10 Android 13
```
chirayudesai, line 4: Nit: I'd leave this lin in

build.gradle
```diff
@@ -1,11 +1,11 @@
...
+    // 1.7.20 Android 13 (TQ2A)
```
chirayudesai: It's usually called QPR2

chirayudesai (dismissed):
We could also rebuild `android.jar` and `libcore.jar`

build.gradle
theimpulson: Done

build.gradle
theimpulson: Thanks, changed

chirayudesai:
I'd like to do a bit of quick testing with this but otherwise it all looks good.

grote:
yeah I'd be most worried about the room upgrade to not affect our K/V DB updates.

theimpulson:
The libs versions seem to be the same in the latest QPR3

theimpulson:
Rebased to fix conflicts

storage/lib/Android.bp
```diff
@@ -40,7 +40,7 @@ android_library {
         enabled: false,
     },
     kotlincflags: [
-        "-Xopt-in=kotlin.RequiresOptIn",
+        "-opt-in=kotlin.RequiresOptIn",
```
grote, line 43: Is this how they do it now? Is that documented somewhere?

gradle/dependencies.gradle
```diff
@@ -1,7 +1,7 @@
 ext {
-    // https://android.googlesource.com/platform/prebuilts/sdk/+/refs/tags/android-13.0.0_r3/current/androidx/Android.bp#3901
-    ext.room_version = "2.4.0-alpha05"
-    // https://android.googlesource.com/platform/external/protobuf/+/refs/tags/android-13.0.0_r3/java/pom.xml#7
+    // https://android.googlesource.com/platform/prebuilts/sdk/+/refs/tags/android-13.0.0_r32/current/androidx/Android.bp#4586
+    ext.room_version = "2.5.0-alpha01"
```
grote: Too bad they include an early alpha for something so critical.

storage/lib/Android.bp
theimpulson: This was hinted at in a compile-time warning. As I see it is mentioned here in docs: https://kotlinlang.org/docs/opt-in-requirements.html#module-wide-opt-in

gradle/dependencies.gradle
theimpulson: Does this matter, as it has been used in AOSP for quite a while?

chirayudesai: It's not used extensively in AOSP.

See https://cs.android.com/search?q=androidx.room_room-runtime&start=1

From what we know, Google mostly checks in prebuilts.

@grote Happy to just check in a prebuilt from this in a separate change (and anything else that we don't want to rely on AOSP versions for - if they version is too old, or alpha, or has any other issue).

theimpulson: Removed for unblocking it, added a comment instead

theimpulson:
Bumped AGP and Gradle to 8.x, ktlint to latest stable version and other changes to accommodate them

grote:
So this forces me to use JDK 17 and AS Giraffe I guess :/

> The project is using an incompatible version (AGP 8.1.0) of the Android Gradle plugin. Latest supported version is AGP 8.0.2

gradle/dependencies.gradle
grote: Thanks for removing. I thought about it and actually there's probably no point in removing it. We build this app with AOSP, so whatever is in AOSP gets used. Removing this here just affects our gradle builds which are dev only.

So, the testing would actually need to happen when AOSP bumps its deps. Ideally, we'd also have some integration tests for K/V backups with room. E.g. Some sample data in the repo and we try to read and compare the result with what we expect. Then just running these tests (on-device!) would already give us more peace of mine.

grote:
So after upgrading my local setup, building it and running tests works here. However, there's an issue with the gradle build light theme where the action bar is all black including the text.

grote (approved):

theimpulson:
> So after upgrading my local setup, building it and running tests works here. However, there's an issue with the gradle build light theme where the action bar is all black including the text.

@grote it seems to be unrelated to this MR. I tried looking and it seems the aforementioned issue is coming from the fact that colors defined for light theme [here](https://github.com/seedvault-app/seedvault/blob/android13/app/src/main/res/values/colors.xml#L7) are using system colors while dark theme colors [here](https://github.com/seedvault-app/seedvault/blob/android13/app/src/main/res/values-night/colors.xml#L7) simply use a properly defined color. I am unsure why it was done that way but maybe we can handle that in a different MR?

gradle/dependencies.gradle
theimpulson: Opened an issue to track things here: https://github.com/seedvault-app/seedvault/issues/548

theimpulson:
> > So after upgrading my local setup, building it and running tests works here. However, there's an issue with the gradle build light theme where the action bar is all black including the text.
> 
> @grote it seems to be unrelated to this MR. I tried looking and it seems the aforementioned issue is coming from the fact that colors defined for light theme [here](https://github.com/seedvault-app/seedvault/blob/android13/app/src/main/res/values/colors.xml#L7) are using system colors while dark theme colors [here](https://github.com/seedvault-app/seedvault/blob/android13/app/src/main/res/values-night/colors.xml#L7) simply use a properly defined color. I am unsure why it was done that way but maybe we can handle that in a different MR?

Issue to keep track: https://github.com/seedvault-app/seedvault/issues/549

PR #551, opened 2023-09-10
Support a lack of MANAGE_DOCUMENTS permission

stevesoltys:
Copy of #480, rebased on `android13`.

This PR is working based on my testing (we may need to address the PR comments in #480 still).

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageActivity.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageActivity.kt
@@ -35,7 +35,9 @@ class StorageActivity : BackupActivity() {
             val authority = uri.authority ?: throw AssertionError("No authority in $uri")
             val storageRoot = StorageRootResolver.getStorageRoots(this, authority).getOrNull(0)
             if (storageRoot == null) {
-                viewModel.onUriPermissionResultReceived(null)
+                viewModel.onSafOptionChosen(
+                    StorageRootResolver.getFakeStorageRootForUri(this, uri))
+                viewModel.onUriPermissionResultReceived(uri)
             } else {
                 viewModel.onSafOptionChosen(storageRoot)
                 viewModel.onUriPermissionResultReceived(uri)
--- a/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageActivity.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageActivity.kt
@@ -31,12 +31,13 @@ class StorageActivity : BackupActivity() {
      */
     private val openDocumentTree = registerForActivityResult(OpenPersistableDocumentTree()) { uri ->
         if (uri != null) {
-            Log.e(TAG, "OpenDocumentTree: $uri")
             val authority = uri.authority ?: throw AssertionError("No authority in $uri")
+            // we are most likely not allowed to resolve storage roots,
+            // but being the optimists we are, we are still trying...
             val storageRoot = StorageRootResolver.getStorageRoots(this, authority).getOrNull(0)
             if (storageRoot == null) {
-                viewModel.onSafOptionChosen(
-                    StorageRootResolver.getFakeStorageRootForUri(this, uri))
+                val fakeRoot = StorageRootResolver.getFakeStorageRootForUri(this, uri)
+                viewModel.onSafOptionChosen(fakeRoot)
                 viewModel.onUriPermissionResultReceived(uri)
             } else {
                 viewModel.onSafOptionChosen(storageRoot)
@@ -58,11 +59,11 @@ class StorageActivity : BackupActivity() {
         }
         viewModel.isSetupWizard = isSetupWizard()
 
-        viewModel.locationSet.observeEvent(this, {
+        viewModel.locationSet.observeEvent(this) {
             showFragment(StorageCheckFragment.newInstance(getCheckFragmentTitle()), true)
-        })
+        }
 
-        viewModel.locationChecked.observeEvent(this, { result ->
+        viewModel.locationChecked.observeEvent(this) { result ->
             val errorMsg = result.errorMsg
             if (errorMsg == null) {
                 setResult(RESULT_OK)
@@ -70,7 +71,7 @@ class StorageActivity : BackupActivity() {
             } else {
                 onInvalidLocation(errorMsg)
             }
-        })
+        }
 
         if (savedInstanceState == null) {
             if (canUseStorageRootsFragment()) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
```diff
@@ -57,6 +58,21 @@ internal object StorageRootResolver {
         return roots
     }
 
+    fun getFakeStorageRootForUri(context: Context, uri: Uri): SafOption {
+        return SafOption(
+            authority = AUTHORITY_STORAGE,
+            rootId = "fake",
+            documentId = "fake",
+            // TODO: Use something other than the USB icon?
+            icon = getIcon(context, AUTHORITY_STORAGE, "usb", 0),
```
grote: Is this shown to the user at some point? If not, we could remove the TODO.

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageActivity.kt
```diff
@@ -35,7 +35,9 @@ class StorageActivity : BackupActivity() {
             val authority = uri.authority ?: throw AssertionError("No authority in $uri")
             val storageRoot = StorageRootResolver.getStorageRoots(this, authority).getOrNull(0)
             if (storageRoot == null) {
-                viewModel.onUriPermissionResultReceived(null)
+                viewModel.onSafOptionChosen(
+                    StorageRootResolver.getFakeStorageRootForUri(this, uri))
+                viewModel.onUriPermissionResultReceived(uri)
```
grote, line 41: I checked and this is in since e9f3c082205d72f873abcdc1d45431839ebf5c83 which introduced support for MANAGE_DOCUMENTS to be optional. So I assume back then, we could get the storage root for the selected authority and now for some reason, this doesn't work anymore, so a fake root is required. I would like to understand why though. Maybe we don't even need a fake root?

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
```diff
@@ -57,6 +58,21 @@ internal object StorageRootResolver {
...
+            title = context.getString(R.string.storage_user_selected_location_title),
+            summary = context.getString(R.string.storage_user_selected_location_summary),
+            availableBytes = null,
+            isUsb = false, // TODO: Check this if possible instead of forcing false
+            requiresNetwork = false, // TODO: Check this if possible instead of forcing false
```
grote: I guess those are the downsides of using a fake root. We can't know how to set those flags.

app/src/main/res/values/strings.xml
```diff
@@ -58,6 +58,8 @@
     <string name="storage_fragment_restore_title">Where to find your backups?</string>
     <string name="storage_fragment_warning">People with access to your storage location can learn which apps you use, but do not get access to the apps\' data.</string>
     <string name="storage_fragment_warning_delete">Existing backups in this location will be deleted.</string>
+    <string name="storage_user_selected_location_title">User-chosen location</string>
+    <string name="storage_user_selected_location_summary">Chosen using the folder browser</string>
```
grote: If this is never shown to the user, we may not need those extra strings.

grote (changes requested):

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageActivity.kt
grote: ok, so we are getting this exception:

``` ruby
 W  Failed to load some roots from com.android.externalstorage.documents
                 java.lang.SecurityException: Permission Denial: reading com.android.externalstorage.ExternalStorageProvider uri content://com.android.externalstorage.documents/root from pid=4106, uid=10164 requires that you obtain access using ACTION_OPEN_DOCUMENT or related APIs
                 	at android.os.Parcel.createExceptionOrNull(Parcel.java:3011)
                 	at android.os.Parcel.createException(Parcel.java:2995)
                 	at android.os.Parcel.readException(Parcel.java:2978)
                 	at android.database.DatabaseUtils.readExceptionFromParcel(DatabaseUtils.java:190)
                 	at android.database.DatabaseUtils.readExceptionFromParcel(DatabaseUtils.java:142)
                 	at android.content.ContentProviderProxy.query(ContentProviderNative.java:481)
                 	at android.content.ContentResolver.query(ContentResolver.java:1219)
                 	at android.content.ContentResolver.query(ContentResolver.java:1151)
                 	at android.content.ContentResolver.query(ContentResolver.java:1107)
                 	at com.stevesoltys.seedvault.ui.storage.StorageRootResolver.getStorageRoots(StorageRootResolver.kt:33)
                 	at com.stevesoltys.seedvault.ui.storage.StorageActivity.openDocumentTree$lambda-0(StorageActivity.kt:36)
                 	at com.stevesoltys.seedvault.ui.storage.StorageActivity.$r8$lambda$zFvojtCBGHPUAkBl4f2e1evcp9I(Unknown Source:0)
                 	at com.stevesoltys.seedvault.ui.storage.StorageActivity$$ExternalSyntheticLambda4.onActivityResult(Unknown Source:4)
                 	at androidx.activity.result.ActivityResultRegistry.doDispatch(ActivityResultRegistry.java:404)
```

which looks like we are no longer allowed to query the root of an Uri we got access to.

grote: Same with Nextcloud:

```ruby
java.lang.SecurityException: Permission Denial: reading com.owncloud.android.providers.DocumentsStorageProvider uri content://org.nextcloud.documents/root from pid=4106, uid=10164 requires that you obtain access using ACTION_OPEN_DOCUMENT or related APIs
```

So it looks like we don't even need to try to query storage roots here.

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
grote: I pushed a new commit that removes the TODO

app/src/main/res/values/strings.xml
grote: I pushed a new commit that removes the summary as the title does seem to get shown to the user.

grote:
@stevesoltys maybe drop the `[UNTESTED]` prefix from the initial commit?

stevesoltys:
Thanks for the updates @grote. Are all of the review comments resolved with your changes? Seems like they might be?

Will update the initial commit as well.

grote:
> Thanks for the updates @grote. Are all of the review comments resolved with your changes? Seems like they might be?

I think they should be. But please review them as well.

> Will update the initial commit as well.

Thanks!

grote (approved):

app/src/main/java/com/stevesoltys/seedvault/ui/storage/StorageRootResolver.kt
stevesoltys: Hmm. I guess we could ask the user if the storage they selected is USB to fix this? Could open a ticket for this if you think it would work...

PR #552, opened 2023-09-10
Add support for local development with an emulator

stevesoltys:
Note that this is based off of the `bugfix/fix-no-manage-documents` PR (forked from #480 so the PRs can be stacked).

We need the base PR's fix to allow usage of the application without `MANAGE_DOCUMENTS` being required.


app/development/DEVELOPMENT.md
theimpulson, line 1: This may be better suited to have on the [wiki](https://github.com/seedvault-app/seedvault/wiki)

stevesoltys: The main benefit I see to version controlling this documentation is that it's coupled alongside these scripts, so it will encourage updating the documentation when we make changes (better coupling of documentation with code changes).

Either way is okay though, if we want to stick with the wiki for all docs I don't mind.

grote: I think it is fine to have in the repo. Could be linked from the wiki. I think the wiki is a git repo internally, so it should have version control as well.

grote (dismissed):
Thanks, great work!

Issue #548, opened 2023-08-03
Have some integration tests for K/V backups with room


theimpulson:
> Thanks for removing. I thought about it and actually there's probably no point in removing it. We build this app with AOSP, so whatever is in AOSP gets used. Removing this here just affects our gradle builds which are dev only.

So, the testing would actually need to happen when AOSP bumps its deps. Ideally, we'd also have some integration tests for K/V backups with room. E.g. Some sample data in the repo and we try to read and compare the result with what we expect. Then just running these tests (on-device!) would already give us more peace of mine.

_Originally posted by @grote in https://github.com/seedvault-app/seedvault/pull/526#discussion_r1282858166_

grote:
@stevesoltys for this, we'd need test data that has a K/V backup and when restoring it, we'd need to compare the data with what the expect. But I guess we could also use app test data and do a backup of that first.

stevesoltys:
Any ideas for apps which produce K/V backups that can be included in the static test data?

grote:
The most important one is this magic package manager sentinel where it stores the app metadata. If we would assert the existence of some of those keys and the hash of some values this would be fine I'd say.

However, for this ticket, ideally, we'd test an AOSP build on a new AOSP emulator image to be sure nothing breaks, but one step after the next.

grote:
@stevesoltys I think your existing tests kinda cover this, right? (We just need to re-enabled them)

PR #553, opened 2023-09-12
Add end-to-end emulator test running on CI

stevesoltys:


app/build.gradle
```diff
@@ -150,10 +157,13 @@ dependencies {
     testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junit5_version"
     testRuntimeOnly "org.junit.vintage:junit-vintage-engine:$junit5_version"
 
+    androidTestImplementation rootProject.ext.aosp_libs
```
stevesoltys, line 160: I think I can remove this...

grote: This *should* already get pulled in via `implementation`.

app/build.gradle
```diff
@@ -21,8 +21,15 @@ android {
         minSdk 32 // leave at 32 for robolectric tests
         targetSdk rootProject.ext.targetSdk
         versionNameSuffix "-$gitDescribe"
-        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        testInstrumentationRunner "com.stevesoltys.seedvault.KoinInstrumentationTestRunner"
```
grote, line 24: Why do we need this? We had instrumentation tests before and they worked, right?

app/build.gradle
```diff
@@ -150,10 +157,13 @@ dependencies {
     testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junit5_version"
     testRuntimeOnly "org.junit.vintage:junit-vintage-engine:$junit5_version"
 
+    androidTestImplementation rootProject.ext.aosp_libs
     androidTestImplementation 'androidx.test:runner:1.4.0'
     androidTestImplementation 'androidx.test:rules:1.4.0'
     androidTestImplementation 'androidx.test.ext:junit:1.1.3'
     androidTestImplementation "io.mockk:mockk-android:$mockk_version"
+
+    androidTestImplementation 'com.kaspersky.android-components:kaspresso:1.5.3'
```
grote: What is this for? Does this get used anywhere?

.idea/runConfigurations/app_emulator.xml
```diff
@@ -62,7 +62,7 @@
     <option name="SEARCH_ACTIVITY_IN_GLOBAL_SCOPE" value="false" />
     <option name="SKIP_ACTIVITY_VALIDATION" value="false" />
     <method v="2">
-      <option name="RunConfigurationTask" enabled="false" run_configuration_name="seedvault [installEmulatorRelease]" run_configuration_type="GradleRunConfiguration" />
+      <option name="Gradle.BeforeRunTask" enabled="false" tasks="installEmulatorRelease" externalProjectPath="$PROJECT_DIR$/app" vmOptions="" scriptParameters="" />
```
grote, line 65: Why is this needed? It wouldn't install the app otherwise?

.cirrus.yml
```diff
@@ -0,0 +1,53 @@
+container:
+  image: ghcr.io/cirruslabs/android-sdk:33
+  kvm: true
+  cpu: 8
+  memory: 16G
+
+check_android_task:
```
grote: Maybe we should give this a more speaking name? `Cirrus CI instrumentation tests`?

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/BackupRestoreTest.kt
```diff
@@ -0,0 +1,139 @@
...
+        verifyCode()
+        chooseBackupLocation()
+
+        val eligiblePackages = launchAllEligibleApps()
+        performBackup(eligiblePackages)
+        uninstallPackages(eligiblePackages)
+        performRestore()
+
+        val packagesAfterRestore = getEligibleApps()
+        assert(eligiblePackages == packagesAfterRestore)
```
grote: What apps are installed on the emulator for testing? Do those apps even have data to get backed up? Do we have a way to assert that the data in the backup really matches what we would expect to get restored?

grote: Related ticket: https://github.com/seedvault-app/seedvault/issues/548

app/build.gradle
stevesoltys: Yeah, this is [to add mockk spies](https://github.com/seedvault-app/seedvault/blob/259ec52054cba4cd5bdbd35832717266829426e6/app/src/androidTest/java/com/stevesoltys/seedvault/KoinInstrumentationTestApp.kt) for some dependencies.

app/build.gradle
stevesoltys: That's [this library](https://github.com/KasperskyLab/Kaspresso) which we use as a [base test class](https://github.com/seedvault-app/seedvault/blob/259ec52054cba4cd5bdbd35832717266829426e6/app/src/androidTest/java/com/stevesoltys/seedvault/e2e/LargeTestBase.kt#L17) and for [some screens](https://github.com/seedvault-app/seedvault/blob/259ec52054cba4cd5bdbd35832717266829426e6/app/src/androidTest/java/com/stevesoltys/seedvault/e2e/screen/impl/RecoveryCodeScreen.kt).

.idea/runConfigurations/app_emulator.xml
stevesoltys: I'd have to check again but I recall using regular `adb install` wasn't sufficient as it would remove it from `priv-app` (I think... I didn't investigate much past it not working).

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/BackupRestoreTest.kt
stevesoltys: Currently [only Etar](https://github.com/seedvault-app/seedvault/blob/259ec52054cba4cd5bdbd35832717266829426e6/.cirrus.yml#L24). Etar will have data because of [this code](https://github.com/seedvault-app/seedvault/blob/259ec52054cba4cd5bdbd35832717266829426e6/app/src/androidTest/java/com/stevesoltys/seedvault/e2e/BackupRestoreTest.kt#L48) which launches every eligible app's activity in an attempt to generate data.

There is no assertion for data just yet but it does validate that the app is re-installed, and the screenshots prove that Etar was successfully backed up and restored - https://cirrus-ci.com/task/5962961425006592

I'm thinking in a subsequent PR we could probably set up a custom AVD `userdata.img` which includes a bunch of apps that have data pre-configured, which we can run this test against and validate data is actually restored (besides viewing the screenshots). Or better yet, maybe we can just restore a Seedvault backup to provision the test emulator instead :wink:

chirayudesai: > Or better yet, maybe we can just restore a Seedvault backup to provision the test emulator instead 😉

A reference backup would be handy for other things too!

We can just manually create it from an emulator install, save the backup itself and entire userdata too, and keep updating it as needed.

stevesoltys: Created a repository for storing these test backups here:

https://github.com/seedvault-app/seedvault-test-data/releases/tag/1

I will work on another PR to start using these

app/build.gradle
stevesoltys: Removed.

app/build.gradle
stevesoltys: Removed this library since we can get by without it and it wasn't providing much value.

.cirrus.yml
stevesoltys: Done.

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/BackupRestoreTest.kt
stevesoltys: We are now using the above as a baseline for the E2E test.

The tests will:
1. Provision the emulator with Seedvault.
2. Download and restore the test backup.
3. Run the test afterwards, with a nice set of data and apps.

It will now also record the entire process and upload the recording as an artifact!

stevesoltys: @grote as of the latest commit we are now also:

1. Intercepting the backup and restore data, hashing and storing it for each package for both K/V and Full.
2. Asserting that the data we intercepted during restore matches the backup data.
3. Asserting that the backup metadata has no `UNKNOWN_ERROR` cases.

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/LargeTestBase.kt
```diff
@@ -0,0 +1,189 @@
...
+        val sharedPreferences = permitDiskReads {
+            PreferenceManager.getDefaultSharedPreferences(targetContext)
+        }
+        sharedPreferences.edit().clear().apply()
+
+        KeyStore.getInstance(ANDROID_KEY_STORE).apply {
+            load(null)
+        }.apply {
+            deleteEntry(KEY_ALIAS_MAIN)
+            deleteEntry(KEY_ALIAS_BACKUP)
```
grote, line 87: do we need to do all this manually or is there another way like resetting/clearing all app-data?

app/build.gradle
```diff
@@ -210,3 +219,14 @@ tasks.register('installEmulatorRelease', Exec) {
...
     }
 }
+
+tasks.register('clearEmulatorAppData', Exec) {
+    group("emulator")
+
+    doFirst {
+        commandLine "${project.projectDir}/development/scripts/clear_app_data.sh"
+
+        environment "ANDROID_SDK_HOME", android.sdkDirectory.absolutePath
```
grote: Is this the official variable name? Here, it is called `ANDROID_HOME`.

app/development/scripts/clear_app_data.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# assert ANDROID_HOME is set
+if [ -z "$ANDROID_SDK_HOME" ]; then
+  echo "ANDROID_SDK_HOME is not set"
```
grote: some variable name question here.

app/development/scripts/clear_app_data.sh
```diff
@@ -0,0 +1,22 @@
...
+  exit 1
+fi
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+DEVELOPMENT_DIR=$SCRIPT_DIR/..
+ROOT_PROJECT_DIR=$SCRIPT_DIR/../../..
+
+EMULATOR_DEVICE_NAME=$($ANDROID_SDK_HOME/platform-tools/adb devices | grep emulator | cut -f1)
```
grote: this limits execution to emulators. is this intentional? why not allow physical test devices?

.idea/runConfigurations/app_emulator.xml
grote: I actually sometimes/often have the issue that Android studio says it installed just fine, but the app wasn't really updated. Maybe this fixes it?

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/screen/impl/DocumentPickerScreen.kt
```diff
@@ -0,0 +1,19 @@
+package com.stevesoltys.seedvault.e2e.screen.impl
+
+import android.widget.EditText
+import com.stevesoltys.seedvault.e2e.screen.UiDeviceScreen
+
+object DocumentPickerScreen : UiDeviceScreen<DocumentPickerScreen>() {
+
+    val createNewFolderButton = findObject { text("CREATE NEW FOLDER") }
+
+    val useThisFolderButton = findObject { text("USE THIS FOLDER") }
```
grote, line 10: ok these are specific to manage documents permission missing, so it won't work on my test phones which isn't a problem right now, but prevent me from running the tests outside of CI (because I can't run emulators).

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/BackupRestoreTest.kt
grote: awesome! 

maybe later, we can also check that certain packages in the test data did get backed up (maybe assert the number of packages as well?) and maybe, if test data is fixed, for some packages compare against expected data hash?

stevesoltys: Yes definitely! All of those would be great improvements. Test data is static as you said so should work.

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/LargeTestBase.kt
stevesoltys: The reason I did it this way is because clearing the app data (with adb) seems to close the app, failing the test.

Since we need to clear the state before each test and after restoring the baseline backup, I couldn't think of a better solution. Agree that it's not the best... open to any ideas :smile:

app/development/scripts/clear_app_data.sh
stevesoltys: The reason I did this was to avoid the person running the tests accidentially targeting their personal phone (like me :laughing:) when running these tests, since it will do some destructive things like uninstalling packages, clearing backups, wiping seedvault, etc.

I can't test on a physical device unfortunately, but if we could think of away around the above could avoid this I'm sure?

.idea/runConfigurations/app_emulator.xml
stevesoltys: Could be yeah! I haven't tested on a real device since I only have one, but it's worth trying!

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/screen/impl/DocumentPickerScreen.kt
stevesoltys: Ahh yeah. Sorry about that. One option is if we can somehow get the [generic system images](https://developer.android.com/topic/generic-system-image/releases) to boot in an emulator, we can test with the permission. I couldn't get this working though :disappointed:. If you know how let me know and we can switch over, it would be better without all the Google apps anyways.

app/build.gradle
stevesoltys: Switched!

app/development/scripts/clear_app_data.sh
grote: We already have some instrumentation tests like this. Do these tests even work on personal phones? Doesn't it require the platform signing key?

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/screen/impl/DocumentPickerScreen.kt
grote: What we could do instead is *not* request the MANAGE_DOCUMENTS permission for instrumentation test (via a modified manifest). So every device would behave the same.

grote (approved):

app/development/scripts/clear_app_data.sh
stevesoltys: It does, but for instance this script in particular could be run against a personal device and clear the Seedvault data. Maybe just need to enforce only one device is connected?

grote: @chirayudesai is there a way to check for a userdebug build device? Maybe that would be more useful than single device?

chirayudesai: `getprop ro.build.type` - user, userdebug, or eng

Also, `ro.build.tags`. `test-keys` for dev, `release-keys` for prod (typically)

Ref: https://dumps.tadiphone.dev/dumps/google/lynx/-/blob/lynx-user-13-TQ3A.230605.012-10204971-release-keys/system/system/build.prop?ref_type=heads

PR #562, opened 2023-09-22
Add experimental support for forcing D2D transfer backups

### LINKED ISSUE #165: Option to back up all apps by pretending to be a device to device transport

stevesoltys:
This adds a toggle in expert settings which enables D2D transfers for backups. Each backup will have a flag in the metadata indicating whether or not it was produced with the D2D transfer toggle enabled.

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataReader.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataReader.kt
@@ -152,7 +152,8 @@ internal class MetadataReaderImpl(private val crypto: Crypto) : MetadataReader {
                 androidVersion = meta.getInt(JSON_METADATA_SDK_INT),
                 androidIncremental = meta.getString(JSON_METADATA_INCREMENTAL),
                 deviceName = meta.getString(JSON_METADATA_NAME),
-                packageMetadataMap = packageMetadataMap
+                d2dBackup = meta.optBoolean(JSON_METADATA_D2D_BACKUP, false),
+                packageMetadataMap = packageMetadataMap,
             )
         } catch (e: JSONException) {
             throw SecurityException(e)
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
@@ -143,12 +143,9 @@ internal class BackupCoordinator(
         @Suppress("UNUSED_PARAMETER") isFullBackup: Boolean,
     ): Boolean {
         val packageName = targetPackage.packageName
-        // Check that the app is not blacklisted by the user
-        val enabled = settingsManager.isBackupEnabled(packageName)
-        if (!enabled) Log.w(TAG, "Backup of $packageName disabled by user.")
-        // We need to exclude the DocumentsProvider used to store backup data.
-        // Otherwise, it gets killed when we back it up, terminating our backup.
-        return enabled && targetPackage.packageName != plugin.providerPackageName
+        val shouldInclude = packageService.shouldIncludeAppInBackup(packageName)
+        if (!shouldInclude) Log.i(TAG, "Excluding $packageName from backup.")
+        return shouldInclude
     }
 
     /**
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
@@ -1,5 +1,6 @@
 package com.stevesoltys.seedvault.transport.backup
 
+import android.app.backup.IBackupManager
 import android.content.Context
 import android.content.pm.ApplicationInfo.FLAG_ALLOW_BACKUP
 import android.content.pm.ApplicationInfo.FLAG_STOPPED
@@ -11,6 +12,7 @@ import android.content.pm.PackageManager
 import android.content.pm.PackageManager.GET_INSTRUMENTATION
 import android.content.pm.PackageManager.GET_SIGNING_CERTIFICATES
 import android.os.RemoteException
+import android.os.UserHandle
 import android.util.Log
 import android.util.Log.INFO
 import androidx.annotation.WorkerThread
@@ -28,11 +30,13 @@ private const val LOG_MAX_PACKAGES = 100
  */
 internal class PackageService(
     private val context: Context,
+    private val backupManager: IBackupManager,
     private val settingsManager: SettingsManager,
     private val plugin: StoragePlugin,
 ) {
 
     private val packageManager: PackageManager = context.packageManager
+    private val myUserId = UserHandle.myUserId()
 
     val eligiblePackages: Array<String>
         @WorkerThread
@@ -48,13 +52,13 @@ internal class PackageService(
                 logPackages(packages)
             }
 
-            // We do not use BackupManager.filterAppsEligibleForBackupForUser because it
-            // always makes its determinations based on OperationType.BACKUP, never based on
-            // OperationType.MIGRATION, and there are no alternative publicly-available APIs.
-            // We don't need to use it, here, either; during a backup or migration, the system
-            // will perform its own eligibility checks regardless. We merely need to filter out
-            // apps that we, or the user, want to exclude.
-            val eligibleApps = packages.filter(::shouldIncludeAppInBackup)
+            val eligibleApps = if (settingsManager.d2dBackupsEnabled()) {
+                // if D2D is enabled, use the "new method" for filtering packages
+                packages.filter(::shouldIncludeAppInBackup).toTypedArray()
+            } else {
+                // otherwise, use the BackupManager call.
+                backupManager.filterAppsEligibleForBackupForUser(myUserId, packages.toTypedArray())
+            }
 
             // log eligible packages
             if (Log.isLoggable(TAG, INFO)) {
@@ -69,6 +73,9 @@ internal class PackageService(
             return packageArray.toTypedArray()
         }
 
+    /**
+     * A list of packages that will not be backed up.
+     */
     val notBackedUpPackages: List<PackageInfo>
         @WorkerThread
         get() {
@@ -97,16 +104,23 @@ internal class PackageService(
     val userApps: List<PackageInfo>
         @WorkerThread
         get() = packageManager.getInstalledPackages(GET_INSTRUMENTATION).filter { packageInfo ->
-            packageInfo.isUserVisible(context) && packageInfo.allowsBackup()
+            packageInfo.isUserVisible(context) &&
+                packageInfo.allowsBackup()
         }
 
     /**
-     * A list of apps that does not allow backup.
+     * A list of apps that do not allow backup.
      */
     val userNotAllowedApps: List<PackageInfo>
         @WorkerThread
-        get() = packageManager.getInstalledPackages(0).filter { packageInfo ->
-            !packageInfo.allowsBackup() && !packageInfo.isSystemApp()
+        get() {
+            // if D2D backups are enabled, all apps are allowed
+            if (settingsManager.d2dBackupsEnabled()) return emptyList()
+
+            return packageManager.getInstalledPackages(0).filter { packageInfo ->
+                !packageInfo.allowsBackup() &&
+                    !packageInfo.isSystemApp()
+            }
         }
 
     val expectedAppTotals: ExpectedAppTotals
@@ -132,11 +146,24 @@ internal class PackageService(
     }
 
     fun shouldIncludeAppInBackup(packageName: String): Boolean {
+        // We do not use BackupManager.filterAppsEligibleForBackupForUser for D2D because it
+        // always makes its determinations based on OperationType.BACKUP, never based on
+        // OperationType.MIGRATION, and there are no alternative publicly-available APIs.
+        // We don't need to use it, here, either; during a backup or migration, the system
+        // will perform its own eligibility checks regardless. We merely need to filter out
+        // apps that we, or the user, want to exclude.
+
         // Check that the app is not excluded by user preference
         val enabled = settingsManager.isBackupEnabled(packageName)
-        // We also need to exclude the DocumentsProvider used to store backup data.
-        // Otherwise, it gets killed when we back it up, terminating our backup.
-        return enabled && packageName != plugin.providerPackageName
+
+        // We need to explicitly exclude DocumentsProvider and Seedvault.
+        // Otherwise, they get killed while backing them up, terminating our backup.
+        val excludedPackages = setOf(
+            plugin.providerPackageName,
+            context.packageName
+        )
+
+        return enabled && !excludedPackages.contains(packageName)
     }
 
     private fun logPackages(packages: List<String>) {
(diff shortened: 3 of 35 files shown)


grote:
We probably need to do #484 cleanly first. Did you not encounter that issue in your tests?

stevesoltys:
Didn't test this just yet but I had done tests with the original D2D branch in the emulator and it didn't have this problem (I produced the static test data backup from that branch). 

I also think it didn't have 300 packages though.

app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataReader.kt
```diff
@@ -152,7 +152,8 @@ internal class MetadataReaderImpl(private val crypto: Crypto) : MetadataReader {
                 androidVersion = meta.getInt(JSON_METADATA_SDK_INT),
                 androidIncremental = meta.getString(JSON_METADATA_INCREMENTAL),
                 deviceName = meta.getString(JSON_METADATA_NAME),
-                packageMetadataMap = packageMetadataMap
+                d2dBackup = meta.optBoolean(JSON_METADATA_D2D_BACKUP, false),
```
grote, line 155: It is trivial, but ideally the `app/src/test/java/com/stevesoltys/seedvault/metadata/MetadataManagerTest.kt` also tests for this.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -45,13 +49,10 @@ internal class PackageService(
...
                 Log.i(TAG, "Got ${packages.size} packages:")
-                packages.chunked(LOG_MAX_PACKAGES).forEach {
-                    Log.i(TAG, it.toString())
-                }
+                logPackages(packages)
             }
 
-            val eligibleApps =
-                backupManager.filterAppsEligibleForBackupForUser(myUserId, packages.toTypedArray())
+            val eligibleApps = packages.filter(::shouldIncludeAppInBackup).toTypedArray()
```
grote: This is done for d2d and normal backup, so doesn't it alter behavior for normal backups?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -45,13 +49,10 @@ internal class PackageService(
...
 
             // log eligible packages
```
grote, line 63: I wonder if many eligible packages will not also cause a binder transaction crash similar to when restoring too many apps.

app/src/main/java/com/stevesoltys/seedvault/transport/restore/RestoreCoordinator.kt
```diff
@@ -92,7 +101,20 @@ internal class RestoreCoordinator(
...
+                D2D_TRANSPORT_FLAGS
+            } else {
+                DEFAULT_TRANSPORT_FLAGS
+            }
+
+            val deviceName = if (metadata.d2dBackup) {
+                D2D_DEVICE_NAME
+            } else {
+                metadata.deviceName
+            }
```
grote, line 115: Is this a user-visible change?

app/src/test/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinatorTest.kt
```diff
@@ -170,20 +168,6 @@ internal class BackupCoordinatorTest : BackupTest() {
...
-        } returns true andThen false andThen true
-        every {
-            plugin.providerPackageName
-        } returns packageInfo.packageName andThen "new.package" andThen "new.package"
-
-        assertFalse(backup.isAppEligibleForBackup(packageInfo, true))
-        assertFalse(backup.isAppEligibleForBackup(packageInfo, true))
-        assertTrue(backup.isAppEligibleForBackup(packageInfo, true))
-    }
-
```
grote, line 186: Why remove this test and not move it to a new place? After all, the code being tested was just moved, right?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -45,13 +49,10 @@ internal class PackageService(
...
             if (Log.isLoggable(TAG, INFO)) {
                 Log.i(TAG, "Got ${packages.size} packages:")
-                packages.chunked(LOG_MAX_PACKAGES).forEach {
-                    Log.i(TAG, it.toString())
-                }
+                logPackages(packages)
             }
 
-            val eligibleApps =
-                backupManager.filterAppsEligibleForBackupForUser(myUserId, packages.toTypedArray())
```
grote, line 54: If we can really remove this, do we still need the `myUserId` and `backupManager` field?

grote:
There's still a DO NOT MERGE commit in this PR. Can we remove that?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
```diff
@@ -143,12 +143,9 @@ internal class BackupCoordinator(
...
         val packageName = targetPackage.packageName
-        // Check that the app is not blacklisted by the user
-        val enabled = settingsManager.isBackupEnabled(packageName)
-        if (!enabled) Log.w(TAG, "Backup of $packageName disabled by user.")
-        // We need to exclude the DocumentsProvider used to store backup data.
-        // Otherwise, it gets killed when we back it up, terminating our backup.
-        return enabled && targetPackage.packageName != plugin.providerPackageName
+        val shouldInclude = packageService.shouldIncludeAppInBackup(packageName)
+        if (!shouldInclude) Log.i(TAG, "Excluding $packageName from backup.")
+        return shouldInclude
```
grote, line 148: Should we also disable quota when d2d is on? I guess it doesn't make sense to apply the non-d2d quota for something that is meant to not have a quota.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -103,7 +105,8 @@ internal class PackageService(
     val userNotAllowedApps: List<PackageInfo>
         @WorkerThread
         get() = packageManager.getInstalledPackages(0).filter { packageInfo ->
-            !packageInfo.allowsBackup() && !packageInfo.isSystemApp()
+            !packageInfo.allowsBackup(settingsManager.d2dBackupsEnabled()) &&
+                !packageInfo.isSystemApp()
```
grote: Can this be non-empty when d2d is on?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
```diff
@@ -157,9 +181,24 @@ internal fun PackageInfo.isSystemApp(): Boolean {
...
+        // So, we return true.
+        // See frameworks/base/services/backup/java/com/android/server/backup/utils/
+        // BackupEligibilityRules.java lines 74-81 and 163-167 (tag: android-13.0.0_r8).
+        true
+    } else {
+        applicationInfo.flags and FLAG_ALLOW_BACKUP != 0
+    }
 }
 
 /**
```
grote, line 229: Check for `FLAG_ALLOW_BACKUP` further down. It may be that the app numbers we show to users in notifications may be off in this PR. Worse, we use this also in other places like when do back up APKs of apps that don't allow backups.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
grote: OK I got 

```
Got 662 packages:
Filtering left 660 eligible packages:
Backup is enabled, request backup...
0/661 - 0% - 
Backup failed: -1000
Backup finished 0/661. Status: -1000
```

grote: It reproducibly fails, but I don't get a log message for why. Most likely the binder transaction, but would be nice to have certainty. Because, if we need to back up in chunks, this complicates things a lot.

grote (changes requested):

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
stevesoltys:
```
// We do not use BackupManager.filterAppsEligibleForBackupForUser for D2D because it
// always makes its determinations based on OperationType.BACKUP, never based on
// OperationType.MIGRATION, and there are no alternative publicly-available APIs.
// We don't need to use it, here, either; during a backup or migration, the system
// will perform its own eligibility checks regardless. We merely need to filter out
// apps that we, or the user, want to exclude.
```

I think this comment is why this was done this way.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
stevesoltys: Nope.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
stevesoltys: Resolved this, good catch.

app/src/main/java/com/stevesoltys/seedvault/transport/restore/RestoreCoordinator.kt
stevesoltys: This I'm not sure, will look at the test videos to see if there's a difference..

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
stevesoltys: IMO, we should launch with quota for now and the user can switch it off if they want with the toggle, instead of making them choose one way or another.

app/src/test/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinatorTest.kt
stevesoltys: Not sure why this one was removed, I've fixed the test.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
stevesoltys: Removed.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
grote: So it is fine if we request apps for backup that the system doesn't want to back up? I feel like it makes us more likely to run into the binder transaction limit, but if we fix that, then that's fine.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
grote: Do you think it is worth to just return empty list then if d2d is on?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
grote: We should test apps with lots of data first, but going over quota means app doesn't get backed up at all. So I'd auto-disable quota when d2d is toggled on, right in the expert settings fragment. can even be a dependency in XML.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
stevesoltys: Yeah that's what I gathered from the comment. Although we can also keep the old logic here if we wanted for the sake of caution.

stevesoltys: I did the above to be safe.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
stevesoltys: I changed the logic - if the user enables D2D backups, unlimited quota will also be enabled automatically.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
stevesoltys: Done.

app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataReader.kt
grote: bump for test

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
grote: Did you push this? Didn't see this in the push diffs github gives me.

grote: Strange, I found it in the overall diff. Looks like some incremental diffs are missing :(
Why not use the XML dependency method instead of manually checking the unlimited quota setting?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
grote: If we don't fix it here, we should create a ticket for it and work on it later.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
stevesoltys: When done this way it doesn't require that D2D is enabled for the user to enable unlimited quota.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
stevesoltys: Added #604.

app/src/main/res/values/strings.xml
```diff
@@ -49,6 +49,8 @@
     <string name="settings_expert_title">Expert settings</string>
     <string name="settings_expert_quota_title">Unlimited app quota</string>
     <string name="settings_expert_quota_summary">Do not impose a limitation on the size of app backups.\n\nWarning: This can fill up your storage location quickly. Not needed for most apps.</string>
+    <string name="settings_expert_d2d_title">Device-to-device backups</string>
+    <string name="settings_expert_d2d_summary">Tell AOSP that our backups are being used for a D2D transfer. This forces backups for most apps, even when they disallow them.\n\nWarning: This is experimental, use at your own risk.\n\nSee more info on our FAQ.</string>
```
grote: Should we warn about the system possibly doing its own non-d2d backups overwriting the last manual d2d backup?

app/src/main/java/com/stevesoltys/seedvault/transport/restore/RestoreCoordinator.kt
grote: any news on the device name?

app/src/test/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinatorTest.kt
```diff
@@ -72,6 +72,22 @@ internal class BackupCoordinatorTest : BackupTest() {
...
+
+        every {
+            packageService.shouldIncludeAppInBackup(packageInfo.packageName)
+        } returns true andThen false andThen true
+
+        every {
+            plugin.providerPackageName
+        } returns packageInfo.packageName andThen "new.package" andThen "new.package"
+
+        assertTrue(backup.isAppEligibleForBackup(packageInfo, true))
```
grote: This was assertFalse before. Why has it changed and is this test still in the spirit of its name?

app/src/main/java/com/stevesoltys/seedvault/transport/restore/RestoreCoordinator.kt
stevesoltys: Nothing user facing that I could see.

app/src/test/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinatorTest.kt
stevesoltys: Moved and updated this test to an integration test because there are problems unit testing `PackageService`.

app/src/main/res/values/strings.xml
stevesoltys: Done, I can't fit any more information there without it being cut off and needing to do some XML layout magic.

grote (approved):

PR #569, opened 2023-09-26
Switch to GitHub actions for E2E tests

stevesoltys:


.github/workflows/test.yml
```diff
@@ -0,0 +1,121 @@
...
+      - '**/*.xml'
+      - '.github/workflows/test.yml'
+
+jobs:
+  instrumentation_tests:
+    runs-on: macos-11
+    strategy:
+      fail-fast: false
+      matrix:
+        android_target: [33, 34]
```
chirayudesai, line 16: I like the idea of testing for 13 in the 14 branch, to make sure any changes we make here don't break stuff / are backwards compatible. As much as possible, at least. We'll try.

app/build.gradle
```diff
@@ -189,7 +199,7 @@ tasks.register('provisionEmulator', Exec) {
     doFirst {
         commandLine "${project.projectDir}/development/scripts/provision_emulator.sh",
                 "seedvault",
-                "system-images;android-33;google_apis;x86_64"
+                "system-images;android-34;google_apis;x86_64"
```
chirayudesai, line 202: But I'd split this into a separate commit so we can easily backport the rest to 13

chirayudesai (dismissed):
Minor nits, but I like, will be really nice to have it all be integrated.

chirayudesai:
We could just merge this to `android14` directly. It'd fail right now, but then #576 should pass.

app/build.gradle
stevesoltys: Done.

chirayudesai (approved):

Issue #549, opened 2023-08-03
Update theme setup for Gradle-only builds


theimpulson:
The current setup is still using `AppCompat` themes as well as unavailable/unresolvable colors that result in UI looking bad on Gradle-only builds. It would be a good idea to refactor them to ensure we use officially available colors/API as well as Material3.

grote:
@theimpulson is this still an issue?

theimpulson:
This has been fixed. Closing the issue.

PR #578, opened 2023-10-09
colors: Switch to public APIs for colors

theimpulson:
## Summary

This PR switches color resources to public APIs instead of system ones. This fixes colors on Gradle builds as well as Android Studio can access them as well.

## Related Issues

- https://github.com/seedvault-app/seedvault/issues/549

## Testing

Compile Gradle build, and check all colors look fine.

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -1,14 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <color name="accent">@*android:color/accent_device_default_light</color>
-    <color name="primary">@*android:color/primary_device_default_settings_light</color>
-    <color name="primaryDark">@*android:color/primary_dark_device_default_settings_light</color>
-    <color name="background">@*android:color/background_device_default_light</color>
-    <color name="actionBarPrimary">@*android:color/primary_device_default_light</color>
-    <color name="statusBarColor">@*android:color/primary_device_default_settings_light
-    </color>
-    <color name="red">@*android:color/error_color_device_default_dark</color>
-    <color name="ic_launcher_background">@*android:color/accent_device_default_light</color>
+    <!-- AOSP colors -->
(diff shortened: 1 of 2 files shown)


theimpulson:
I added comments similar to dependencies in Gradle files to keep track of source colors. Re-used colors that were the same.

chirayudesai:
In theory the actions runs should show have screenshots / videos to verify some parts.

chirayudesai:
@stevesoltys any ideas why this didn't have the other actions workflows? The emulator tests.

stevesoltys:
@chirayudesai Should be fixed with #580.

chirayudesai:
@theimpulson could you rebase?

app/src/main/res/values/colors.xml
```diff
@@ -1,14 +1,25 @@
...
-    <color name="primaryDark">@*android:color/primary_dark_device_default_settings_light</color>
-    <color name="background">@*android:color/background_device_default_light</color>
-    <color name="actionBarPrimary">@*android:color/primary_device_default_light</color>
-    <color name="statusBarColor">@*android:color/primary_device_default_settings_light
-    </color>
-    <color name="red">@*android:color/error_color_device_default_dark</color>
-    <color name="ic_launcher_background">@*android:color/accent_device_default_light</color>
+    <!-- AOSP colors -->
+    <!-- https://android.googlesource.com/platform/frameworks/base/+/refs/tags/android-14.0.0_r1/core/res/res/values/colors_device_defaults.xml#35 -->
+    <color name="accent">@android:color/system_accent1_600</color>
```
grote, line 5: How does this work when the accent color is changed in theme picker or through wallpaper? Also, what's the difference to that `@*android:` way of specifying colors?

theimpulson: It uses a new accent color then. `@*android:` is resolved at runtime if I remember correctly and the said accent colors are system only.

Reference:

- [R.color](https://developer.android.com/reference/android/R.color)
- [Dynamic color](https://m3.material.io/styles/color/dynamic-color/user-generated-color)

grote (approved):
code changes look good to me

theimpulson:
Seems fine to me on both AOSP & Gradle builds.

PR #581, opened 2023-10-10
Version upgrades of included libraries and test dependencies

### LINKED ISSUE #564: Port to Android 14
* [x] Build Android and Java libraries from AOSP source code that include the hidden APIs required for Seedvault and add them to the (gradle) buildsystem
* [x]  Identify all upgraded libraries and find matching versions for the non-AOSP buildsystem, bumping their versions as needed.
* [x] Upgrade other libraries and include jar binaries as needed.
* [x]  Go through all backup related AOSP commits since last Android release and learn about changes, adapting Seedvault code as needed, especially for breaking changes preventing compilation.
* [x]  Include all work in a new pull request creating a new android14 branch of Seedvault.

Related to https://github.com/seedvault-app/seedvault/issues/433

grote:
These are not strictly needed for Android, but historically, we are using the opportunity when testing new Android versions in depth to also bump dependency versions. Otherwise, we'd never to it.

Note that this intentionally is not upgrading koin as the plan is to replace it with hilt rather sooner than later.

related to #564


chirayudesai:
>  Otherwise, we'd never to it.

I'm happy to have the upgrade done at the same time, but shipping it in production could be a bit later to avoid any potential issues found here from blocking shipping.

chirayudesai:
> for Android 14

Drop this from title given they aren't needed for 14

stevesoltys (approved):

PR #577, opened 2023-10-08
[13] Switch to GitHub actions for E2E tests

stevesoltys:
#569 but for `android-13`.

PR #582, opened 2023-10-11
Fix and improve E2E tests

stevesoltys:


grote:
How does this relate to https://github.com/seedvault-app/seedvault/pull/577 ?

stevesoltys:
@grote it'll likely replace most of it

grote:
Slightly unrelated: we should probably update the status badge in the README.

.github/workflows/test.yml
```diff
@@ -31,88 +30,33 @@ jobs:
...
+              ./.github/scripts/run_tests.sh
+
+      - name: Upload test results
         if: always()
         uses: actions/upload-artifact@v3
         with:
-          name: seedvault_test_videos
-          path: seedvault_test_videos/**/*.mp4
+          name: ${{ matrix.emulator_type }}-${{ matrix.android_target }}-results
+          path: seedvault_test_results/**/*
```
grote, line 62: Is this uploading logcat as well?

.github/workflows/test.yml
```diff
@@ -31,88 +30,33 @@ jobs:
...
-          script: |
-            ./app/development/scripts/provision_emulator.sh "test" "system-images;android-${{ matrix.android_target }};${{ matrix.emulator_type }};x86_64"
-            echo "Generated AVD snapshot for caching."
-
       - name: Assemble tests
         run: ./gradlew :app:assembleAndroidTest
 
       - name: Run tests
-        uses: reactivecircus/android-emulator-runner@v2
+        uses: Wandalen/wretry.action@v1.3.0
```
grote, line 40: we need to retry 3 times? Are things this flaky? do we know where the flakyness comes from?

grote (approved):

.github/workflows/test.yml
stevesoltys: Yeah.

.github/workflows/test.yml
stevesoltys: I'm not sure if we do, just leaving this for now. If we see consistency we can remove it, if not we can fix the flakiness.

PR #585, opened 2023-10-13
Switch to Kotlin build scripts + version catalog

stevesoltys:


build.libs.toml
```diff
@@ -0,0 +1,82 @@
...
+junit4 = "4.13.2"
+junit5 = "5.10.0" # careful, upgrading this can change a Cipher's IV size in tests!?
+mockk = "1.13.4" # newer versions require kotlin > 1.8.10
+espresso = "3.4.0"
+
+# Dependency versions below this are AOSP versions.
+# We use "strictly" to enforce the version cannot be overriden by transitive dependencies.
+# We need to enforce that the versions we use are the same as AOSP to ensure compatibility.
+
+# Kotlin versions
```
chirayudesai, line 19: The comment for this got lost, https://android.googlesource.com/platform/external/kotlinc/+/refs/tags/android-14.0.0_r1/build.txt

build.plugins.toml
```diff
@@ -0,0 +1,5 @@
+[versions]
+kotlin = "1.8.10"
```
grote, line 10: should probably keep the comments for the various kotlin versions and why we are using this one, here.

storage/lib/build.gradle.kts
```diff
@@ -0,0 +1,107 @@
...
+    kotlin("android")
+}
+
+android {
+    namespace = "org.calyxos.backup.storage"
+    compileSdk = libs.versions.compileSdk.get().toInt()
+
+    defaultConfig {
+        minSdk = libs.versions.minSdk.get().toInt()
+        targetSdk = libs.versions.targetSdk.get().toInt()
```
grote: looks like this is deprecated, probably because we don't need to set this for libraries.

There's some other deprecation warnings (e.g. lintOptions -> lint), maybe already address them in an extra commit?

build.gradle.kts
```diff
@@ -0,0 +1,44 @@
+import org.jlleitschuh.gradle.ktlint.KtlintExtension
+
+buildscript {
+    repositories {
+        google()
+        jcenter()
```
grote: do we still need jcenter?

build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+    id("org.jetbrains.dokka") version plugins.versions.kotlin apply false
+    id("org.jlleitschuh.gradle.ktlint") version plugins.versions.ktlint apply false
+}
+
+tasks.register("clean", Delete::class) {
+    delete(rootProject.buildDir)
+}
+
+subprojects {
+    if (path == ":app" || path == ":storage:lib") {
```
grote: I guess we *could* enable this everywhere?

grote (commented):
Awesome work, thanks a lot Steve!

build.gradle.kts
stevesoltys: Enabled for all except `demo`, which has a few violations.

build.gradle.kts
stevesoltys: Removed.

storage/lib/build.gradle.kts
stevesoltys: Removed and fixed most others.

build.plugins.toml
stevesoltys: Added the AOSP link here. Any other comments you think would be good here let me know and will add it.

build.libs.toml
stevesoltys: Added it back.

build.plugins.toml
grote: I found the comments we had before helpful, explaining that we need to match kotlin versions of AOSP here and what Android version was using which Kotlin version, but I guess we can build this up again starting from 14.

grote (dismissed):

PR #586, opened 2023-10-16
Build Seedvault with AOSP in CI

stevesoltys:


.github/scripts/build_aosp.sh
```diff
@@ -0,0 +1,85 @@
...
+export DISABLE_ROBO_RUN_TESTS=true
+
+# set git identity
+git config --global user.email "seedvault@nowhere.com"
+git config --global user.name "Seedvault"
+
+retry curl https://storage.googleapis.com/git-repo-downloads/repo > ~/bin/repo
+chmod a+x ~/bin/repo
+
+# create custom manifest with Seedvault included
```
chirayudesai: We don't need to do a modification like this, there can be a

`.repo/local_manifests/seedvault.xml` created after repo init with the repo.

Also, I'd fetch it to `packages/apps/Seedvault`

.cirrus.yml
```diff
@@ -0,0 +1,12 @@
+task:
+  name: seedvault-aosp
+  timeout_in: 120m
+  container:
+    image: ubuntu:23.04
+    cpu: 8
+    memory: 32G
+  build_script:
+    - ./.github/scripts/build_aosp.sh aosp_arm eng
```
chirayudesai: Should be aosp_arm64

.github/scripts/build_aosp.sh
```diff
@@ -0,0 +1,85 @@
...
+cd /aosp-manifest
+seedvault_manifest="<remote name=\"seedvault\" fetch=\"https://github.com/$CIRRUS_REPO_OWNER/\" revision=\"$CIRRUS_CHANGE_IN_REPO\" />\n  <project path=\"external/seedvault\" name=\"$CIRRUS_REPO_NAME\" remote=\"seedvault\" />"
+sed -i "/<manifest>/a ${seedvault_manifest}" "default.xml"
+sed -i 's/fetch=".."/fetch="https:\/\/android.googlesource.com\/"/g' "default.xml"
+git add .
+git commit -m "Add Seedvault"
+
+mkdir -p /aosp
+cd /aosp
+retry yes | repo init -u /aosp-manifest -b main --depth=1
```
chirayudesai: `main` is wrong - that's the development branch, but given Android is developed privately partially, for frameworks stuff it's a mess.

Should make this customisable, and for now could default to

`android-14.0.0_r1` for seedvault `android14` branch.
`android-13.0.0_r32` for `android13`.


Also note that the tags aren't linear, so we can't always just pick the latest `rX` tag. Both of these tags are in our `gradle/dependencies.gradle` though (and we keep that updated)

.github/scripts/build_aosp.sh
```diff
@@ -0,0 +1,85 @@
...
+  fi
+
+  set -e
+
+  return $exitCode
+}
+
+DEVICE=$1
+TARGET=$2
+export DISABLE_ROBO_RUN_TESTS=true
```
chirayudesai: Shouldn't be needed

.github/scripts/build_aosp.sh
```diff
@@ -0,0 +1,85 @@
...
+
+# set git identity
+git config --global user.email "seedvault@nowhere.com"
```
chirayudesai: Pedantic: that looks like a real site, I'd rather seedvault.app or example.com

.github/scripts/build_aosp.sh
```diff
@@ -0,0 +1,85 @@
...
+git config --global user.name "Seedvault"
```
chirayudesai: Pedantic: Maybe SeedVault CI?

.github/scripts/build_aosp.sh
```diff
@@ -0,0 +1,85 @@
...
+seedvault_manifest="<remote name=\"seedvault\" fetch=\"https://github.com/$CIRRUS_REPO_OWNER/\" revision=\"$CIRRUS_CHANGE_IN_REPO\" />\n  <project path=\"external/seedvault\" name=\"$CIRRUS_REPO_NAME\" remote=\"seedvault\" />"
+sed -i "/<manifest>/a ${seedvault_manifest}" "default.xml"
+sed -i 's/fetch=".."/fetch="https:\/\/android.googlesource.com\/"/g' "default.xml"
+git add .
+git commit -m "Add Seedvault"
+
+mkdir -p /aosp
+cd /aosp
+retry yes | repo init -u /aosp-manifest -b main --depth=1
+retry repo sync -c -j6 --fail-fast --force-sync
```
chirayudesai: Can go upto j8 here

.github/scripts/build_aosp.sh
```diff
@@ -0,0 +1,85 @@
...
+
+# Cirrus CI seems to (possibly) reschedule tasks that aren't sending out logs for a while?
+while true; do echo "Still building..."; sleep 30; done &
+
+source build/envsetup.sh
+lunch $DEVICE-$TARGET
+m -j6 Seedvault
```
chirayudesai, line 89: This can just be `m Seedvault`

chirayudesai (changes requested):

.cirrus.yml
```diff
@@ -0,0 +1,12 @@
...
+  timeout_in: 120m
+  container:
+    image: ubuntu:23.04
+    cpu: 8
+    memory: 32G
+  build_script:
+    - ./.github/scripts/build_aosp.sh aosp_arm64 eng
+  always:
+    seedvault_artifacts:
+      path: "/aosp/out/target/product/generic_arm64/system/priv-app/Seedvault/Seedvault.apk"
```
stevesoltys:
```
out/target/product/generic_arm64/system/system_ext/priv-app/Seedvault/Seedvault.apk
```

.github/scripts/build_aosp.sh
chirayudesai:
```xml
repo init -u https://android.googlesource.com/platform/manifest -b $BRANCH --depth=1
mkdir .repo/local_manifests
cat << EOF > .repo/local_manifests/seedvault.xml
<?xml version="1.0" encoding="UTF-8"?>
<manifest>

  <remote name="seedvault"
          fetch="https://github.com/seedvault-app" />

  <project name="seedvault" path="packages/apps/Seedvault" revision="android14" remote="seedvault" />

</manifest>
EOF
repo sync -c -j8 --fail-fast --force-sync
```

.github/scripts/build_aosp.sh
chirayudesai: Or not, need to stick with lower jobs on resource-constrained machines.

.cirrus.yml
```diff
@@ -0,0 +1,12 @@
+task:
+  name: seedvault-aosp
+  timeout_in: 120m
+  container:
+    image: ubuntu:23.04
+    cpu: 8
+    memory: 32G
+  build_script:
+    - ./.github/scripts/build_aosp.sh aosp_arm64 eng android-14.0.0_r1
```
chirayudesai: Any reason to go with `eng` instead of `userdebug` or just `user` even?

For an app like this it shouldn't make too big of a runtime difference. I'd default to `userdebug`.

chirayudesai (commented):
Looks good to me, thanks!

.cirrus.yml
stevesoltys: Sure. Will add this to my other PR to avoid wasting build minutes.

chirayudesai: Could also just cancel the build right after push. :)

PR #589, opened 2023-10-27
Allowlist ProtonDrive

### LINKED ISSUE #432: Add support for Proton Drive
Please add support for Proton Drive (currently in Beta for Android)

chirayudesai:
Fixes: #432

### DIFF
--- a/app/src/debug/res/values/config.xml
+++ b/app/src/debug/res/values/config.xml
@@ -7,5 +7,6 @@
         <item>org.nextcloud.documents</item>
         <item>org.nextcloud.beta.documents</item>
         <item>at.bitfire.davdroid.webdav</item>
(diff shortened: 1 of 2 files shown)


chirayudesai:
Untested, but I wanted CI to a build an apk for quick testing

Forage:
Thanks @chirayudesai. You missed https://github.com/seedvault-app/seedvault/blob/android14/app/src/debug/res/values/config.xml though.

grote:
If the protonDrive app doesn't support open document tree, we can close this.

chirayudesai:
Unfortunately it looks like we won't be able to add it because the ProtonDrive app does not support `ACTION_OPEN_DOCUMENT_TREE` - which is what we need to store all the backup files.

PR #591, opened 2023-11-01
fixup! Allow secondary user backup to USB

Uldiniad:
Issue: [calyxos#1058](https://gitlab.com/CalyxOS/calyxos/-/issues/1058)
Change-Id: I4aaadef5954e9c091f7048bffd8e0ed700fdd2e6

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/App.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/App.kt
@@ -5,12 +5,11 @@ import android.app.Application
 import android.app.backup.BackupManager.PACKAGE_MANAGER_SENTINEL
 import android.app.backup.IBackupManager
 import android.content.Context
-import android.content.Context.BACKUP_SERVICE
 import android.content.pm.PackageManager.PERMISSION_GRANTED
 import android.os.Build
 import android.os.ServiceManager.getService
 import android.os.StrictMode
-import android.os.UserHandle
+import android.os.UserManager
 import com.stevesoltys.seedvault.crypto.cryptoModule
 import com.stevesoltys.seedvault.header.headerModule
 import com.stevesoltys.seedvault.metadata.MetadataManager
@@ -143,8 +142,11 @@ fun <T> permitDiskReads(func: () -> T): T {
     }
 }
 
-fun Context.getSystemContext(isUsbStorage: () -> Boolean): Context {
-    return if (checkSelfPermission(INTERACT_ACROSS_USERS_FULL) == PERMISSION_GRANTED &&
-        isUsbStorage()
-    ) createContextAsUser(UserHandle.SYSTEM, 0) else this
+@Suppress("MissingPermission")
+fun Context.getStorageContext(isUsbStorage: () -> Boolean): Context {
+    if (checkSelfPermission(INTERACT_ACROSS_USERS_FULL) == PERMISSION_GRANTED && isUsbStorage()) {
+        UserManager.get(this).getProfileParent(user)
+            ?.let { parent -> return createContextAsUser(parent, 0) }
+    }
+    return this
 }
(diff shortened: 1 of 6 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -55,6 +55,9 @@
     <uses-permission android:name="android.permission.USE_BIOMETRIC" />
 
     <!-- This is needed to query content providers in other users -->
+    <uses-permission
+        android:name="android.permission.INTERACT_ACROSS_USERS"
```
chirayudesai: `_FULL` which we already have below is defined as the fuller form of this, so it's fine.

Uldiniad: Lint complained when I didn't include this

app/src/main/java/com/stevesoltys/seedvault/App.kt
```diff
@@ -143,8 +142,10 @@ fun <T> permitDiskReads(func: () -> T): T {
     }
 }
 
-fun Context.getSystemContext(isUsbStorage: () -> Boolean): Context {
-    return if (checkSelfPermission(INTERACT_ACROSS_USERS_FULL) == PERMISSION_GRANTED &&
-        isUsbStorage()
-    ) createContextAsUser(UserHandle.SYSTEM, 0) else this
+fun Context.getContext(isUsbStorage: () -> Boolean): Context {
```
chirayudesai: Maybe `getParentContext` ? Or something along that lines.

The main thing last time was to avoid confusion with the typically used `getContext` everywhere in Android, which this is not.

chirayudesai (changes requested):

app/src/main/AndroidManifest.xml
chirayudesai: Yes I meant adding this permission is fine if needed, but I'd just ignore Lint if it actually works.

app/src/main/java/com/stevesoltys/seedvault/App.kt
Uldiniad: This function takes a boolean as a parameter though. Otherwise, parent context would make it seem like we're always obtaining that context but that's not always applicable.

app/src/main/AndroidManifest.xml
Uldiniad: Sure, but someone will have to merge while bypassing it since otherwise it shows build failed below

chirayudesai: No I mean you'd actually have it ignore it, by an override

app/src/main/java/com/stevesoltys/seedvault/App.kt
chirayudesai: My point here was it should be named something other than plain `getContext`, because this isn't that.

How about `getActualOrParentContext`

Uldiniad: changed to `getStorageContext`

app/src/main/AndroidManifest.xml
Uldiniad: Added `@Supress`

chirayudesai (approved):

grote (commented):
Code looks good. Can we get some more explanation what this is about and what the issue is this fixes? Why did this work before? Is this a work-profile vs. secondary user thing? Or is this Android 13 vs. Android 14?

Uldiniad:
> Code looks good. Can we get some more explanation what this is about and what the issue is this fixes? Why did this work before? Is this a work-profile vs. secondary user thing? Or is this Android 13 vs. Android 14?

Issue it fixes is linked in the description. If the USB is plugged in beforehand and remain plugged in when creating and switching to a new full secondary user, the current code works because the USB is mounted in user 0. But if the USB is plugged in after switching to the new user, USB is mounted in it instead. Effectively, this latter case was never working but was not previously caught. This does not affect work profiles (at least those in the main user since the foreground user is the same as the parent of the work profile). This bug applies no matter the Android version

PR #595, opened 2023-11-20
Support installing via F-Droid Basic

t-m-w:
If an app being restored was unable to be installed from a backup, e.g. because it was missing, and the app was originally installed with F-Droid Basic, try to install it using F-Droid Basic again, as is done with other listed installers.

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/restore/install/InstallIntentCreator.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/restore/install/InstallIntentCreator.kt
@@ -14,6 +14,7 @@ internal class InstallIntentCreator(
     private val installerToPackage = mapOf(
         "org.fdroid.fdroid" to "org.fdroid.fdroid",
         "org.fdroid.fdroid.privileged" to "org.fdroid.fdroid",
+        "org.fdroid.basic" to "org.fdroid.basic",
         "com.aurora.store" to "com.aurora.store",
         "com.aurora.services" to "com.aurora.store",
         "com.android.vending" to "com.android.vending"


t-m-w:
What this change doesn't try to do (and I haven't tested the behavior) is ensure that F-Droid Basic is actually available before reaching this point.

From discussion, if I'm understanding right, this only becomes relevant if some apps fail to be restored (or APK backup was turned off, so the APKs aren't available). It requires the user to tap to install the app, which uses the logic here to decide which installer app to launch.

This could be a potential problem for any of the installers, I think, though I'm not sure what if anything could be done about it, at least on the restoration side of Seedvault.

app/src/main/java/com/stevesoltys/seedvault/restore/install/InstallIntentCreator.kt
```diff
@@ -14,6 +14,7 @@ internal class InstallIntentCreator(
     private val installerToPackage = mapOf(
         "org.fdroid.fdroid" to "org.fdroid.fdroid",
         "org.fdroid.fdroid.privileged" to "org.fdroid.fdroid",
+        "org.fdroid.basic" to "org.fdroid.basic",
```
chirayudesai, line 17: Just out of curiosity - what would happen if it wasn't on this list? I only ask for the 1:1 mapping such as this one.

chirayudesai:
> 
> This could be a potential problem for any of the installers, I think, though I'm not sure what if anything could be done about it, at least on the restoration side of Seedvault.

Yes I don't think we can solve it here. Anyone including SeedVault would have to solve it in however way they see fit - and there's multiple ways to do it.

Most we can do is show helpful error messages to the user.

chirayudesai (approved):

app/src/main/java/com/stevesoltys/seedvault/restore/install/InstallIntentCreator.kt
t-m-w: As far as I can tell, it would return a null intent (nothing to install the app with), but as far as the user experience, I'd have to look further.

grote:
> What this change doesn't try to do (and I haven't tested the behavior) is ensure that F-Droid Basic is actually available before reaching this point.

The way I am reading the code is that if the installer is not installed itself, we are are opening up the intent for other `market://` intent handling apps to grab. So the user should see a chooser if more than one such app is available.

grote (approved):

PR #603, opened 2024-01-03
Stop backing up excluded app APKs

### LINKED ISSUE #594: APKs of apps that don't allow backup do get backed up even if APK backup is off
We may be doing this, so the user can get these apps back. They may not appear in the list of apps to manually install. However, for many apps that don't allow backup, this could lead to significant storage requirements and cause time-outs.

If the reasoning is true, we should store the list of those apps and just add them to the list of apps to restore, even if we don't have data for them.

stevesoltys:
Resolves #594

### DIFF
--- a/app/src/test/java/com/stevesoltys/seedvault/transport/backup/ApkBackupTest.kt
+++ b/app/src/test/java/com/stevesoltys/seedvault/transport/backup/ApkBackupTest.kt
@@ -62,6 +62,15 @@ internal class ApkBackupTest : BackupTest() {
     @Test
     fun `does not back up when setting disabled`() = runBlocking {
         every { settingsManager.backupApks() } returns false
+        every { settingsManager.isBackupEnabled(any()) } returns true
+
+        assertNull(apkBackup.backupApkIfNecessary(packageInfo, UNKNOWN_ERROR, streamGetter))
+    }
+
+    @Test
+    fun `does not back up when app blacklisted`() = runBlocking {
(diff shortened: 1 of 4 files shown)


grote (dismissed):
Looks good. Need to look into why tests don't pass.

app/src/test/java/com/stevesoltys/seedvault/transport/backup/ApkBackupTest.kt
```diff
@@ -53,6 +54,12 @@ internal class ApkBackupTest : BackupTest() {
         mockkStatic(PackageUtils::class)
     }
 
+    @BeforeEach
+    fun setUp() {
+        every { settingsManager.backupApks() } returns true
+        every { settingsManager.isBackupEnabled(any()) } returns true
+    }
```
grote: For small things like these, I'd prefer to have them explicitly in each test, especially when the test outcome depends on it, e.g. `settingsManager.backupApks()`

stevesoltys: I would argue that establishing a default state for a "happy path" and then having tests which deviate from that path makes the tests easier to understand and reduces duplicate code.

This is subjective though so I've reverted the change.

grote: Thanks!

grote (approved):

PR #605, opened 2024-01-16
Store and show the size of app backups

grote:
This PR has logically split commits. Storing of the size is one commit and exposing that size in the UI is other commits. A review may want to look at commits individually.


grote:
Help with testing https://github.com/seedvault-app/seedvault/issues/165

grote:
We could also show the size in other places like when restoring, but for now this is good enough.

stevesoltys (approved):

PR #613, opened 2024-02-06
Use BackupRequester to request backup in chunks

### LINKED ISSUE #604: Binder exception when backing up a large number of applications
As exposed by #562, and similar to the issue resolved for restore in #563, backing up a large number of applications leads to a binder exception. This can be resolved by backing up in chunks, as we did in #563.

grote:
Draft, because based on scheduling PR, but ready for review.

Fixes #604

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupRequester.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupRequester.kt
@@ -38,7 +38,7 @@ internal class BackupRequester(
     private val observer = NotificationBackupObserver(
         context = context,
         backupRequester = this,
-        expectedPackages = packages.size,
+        requestedPackages = packages.size,
         appTotals = packageService.expectedAppTotals,
     )
     private val monitor = BackupMonitor()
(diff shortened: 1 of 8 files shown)


grote:
@stevesoltys can you help me with the e2e test here that I still can't run myself locally? It seems the @pm@ data changes when requesting backup in chunks.

stevesoltys (approved):

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupRequester.kt
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * Used for splitting the packages into chunks.
+     */
+    private var packageIndex: Int = 0
+
+    /**
+     * Request the backup to happen. Should be called short after constructing this object.
+     */
+    fun requestBackup(): Boolean {
+        if (packageIndex != 0) error("requestBackup() called more than once!")
```
stevesoltys, line 57: Should this return false instead of throwing an exception?

grote: I am a fan of throwing exceptions when the assumptions we make about how this class gets used by our own code is violated. If we return false, this will be normal operation and we never notice that we use it wrong.

Issue #336, opened 2021-10-30
Storage backup bugs


rugk:
## What happened

I noticed that since some days my backup would not show as backed up completely, i.e. the day counter showed "last backup: 21 October", e.g.
Also, I noticed that once a day the beat feature for storage backup starts and does sth., which I interpret as being successful.

Now, I thus triggered it manually and I see it seems to do some backup and even pruning then, but at the end it shows a failed notification:
![backup-failed](https://user-images.githubusercontent.com/11966684/139514792-a9554f11-4630-4fb5-8e92-dcca08fcead0.png)

I retriggered it 2-3 times with the same result…

## Notable things

* The menu of Seedvault shows a red icon next to the app "Nextcloud" indicating that the Nexctcloud app could not be backed up. Maybe the Nextcloud data could not be backed up while Nextcloud was currently being used for uploading (other) backup data?
* As it shows something about pruning, maybe actually pruning fails and not the backup?
  * If that is the case, and sth. was actually backed up, maybe the status display in the app is wrong though, because it still shows some apps waiting to be backed up, which are new and likely included in the backup now. So it seems if pruning fails, the status is not updated, even though the backup may have suceeded.

### Some logcat snippets

I can **provide the full logcat privately** if needed.

Here are some interesting snippets I could find upon skimming it:

<details>

```
[10-29 13:46:38.740 18678:18709 W/UnsafeUtil]
platform method missing - proto runtime falling back to safer methods: java.lang.NoSuchMethodException: sun.misc.Unsafe.copyMemory [class java.lang.Object, long, class java.lang.Object, long, long]

[10-29 13:46:38.755 3287:3306 I/MediaProvider]
Using lower FS for /storage/emulated/0/Download/blokada

[10-29 13:46:38.763 18678:18709 E/SmallFileBackup]
Error backing up content://media/external/downloads/3551
java.io.IOException: read failed: EISDIR (Is a directory)
	at libcore.io.IoBridge.read(IoBridge.java:519)
	at java.io.FileInputStream.read(FileInputStream.java:320)
	at android.os.ParcelFileDescriptor$AutoCloseInputStream.read(ParcelFileDescriptor.java:1005)
	at java.io.FileInputStream.read(FileInputStream.java:289)
	at android.os.ParcelFileDescriptor$AutoCloseInputStream.read(ParcelFileDescriptor.java:996)
	at kotlin.io.ByteStreamsKt.copyTo(IOStreams.kt:106)
[...]

PR #619, opened 2024-02-12
Don't include empty dirs in media backup

grote:
Otherwise, the directories would be treated as small files and we would try to backup their content which would fail.
Non-empty directories still get restored when we restore the files in them.

How to test: Tell Seedvault to backup Downloads, then use file manager to create a folder inside `/sdcard/Download`.

Fixes one bug of #336

### DIFF
--- a/storage/lib/src/main/java/org/calyxos/backup/storage/scanner/MediaScanner.kt
+++ b/storage/lib/src/main/java/org/calyxos/backup/storage/scanner/MediaScanner.kt
@@ -12,6 +12,7 @@ import android.database.Cursor
 import android.net.Uri
 import android.os.Bundle
 import android.os.Environment
+import android.provider.DocumentsContract.Document.MIME_TYPE_DIR
 import android.provider.MediaStore
 import android.provider.MediaStore.MediaColumns.IS_DOWNLOAD
 import android.util.Log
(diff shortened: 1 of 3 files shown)


storage/lib/src/main/java/org/calyxos/backup/storage/scanner/MediaScanner.kt
```diff
@@ -57,15 +58,18 @@ public class MediaScanner(context: Context) {
 
     internal fun scanMediaUri(uri: Uri, extraQuery: String? = null): List<MediaFile> {
         val extras = Bundle().apply {
-            val query = StringBuilder()
+            val query = StringBuilder().apply {
```
stevesoltys, line 61: Tests?

stevesoltys (approved):

storage/lib/src/main/java/org/calyxos/backup/storage/scanner/MediaScanner.kt
grote: we don't have tests for this part, because we'd need a phone image with known files to test against. maybe we can make a ticket for you to provide these files as part of the image or just let me know where they are? (still can't run the isntrumentation tests locally)

PR #620, opened 2024-02-12
UNTESTED: Don't backup device or profile owner apps

grote:


grote:
We can close this, because CalyxOS fixed this ROM-side.

Issue #590, opened 2023-10-27
Backup of many APKs time out and make backup fail


azarok:
I spoke with Nick from Clyxos, and he told me to provide you the log.
There are some personal details in the log. How might I attach the file such that no one a part of you could view it?
Thank you

grote:
@chirayudesai didn't we have an email address for these cases?

chirayudesai:
We have one for calyx, bugreports@calyxos.org

IIRC GitHub doesn't allow others to open private / confidential issues, otherwise that'd have been another way.

azarok:
Thank you both, I have just emailed you with the log.

t-m-w:
The backup proceeds successfully for a while but eventually encounters a particular package, stalls, and times out. (I removed anything even close to being identifiable from this snippet.)

```
I BackupNotificationManager: [redacted progress] - Opt-out APK for [redacted package]
[ ~10 seconds later ]
W BackupTransportClient: Failed to get result from transport:
W BackupTransportClient: java.util.concurrent.TimeoutException
W BackupTransportClient: 	at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1862)
W BackupTransportClient: 	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1997)
W BackupTransportClient: 	at com.android.server.backup.transport.BackupTransportClient.getFutureResult(BackupTransportClient.java:377)
W BackupTransportClient: 	at com.android.server.backup.transport.BackupTransportClient.requestBackupTime(BackupTransportClient.java:159)
W BackupTransportClient: 	at com.android.server.backup.keyvalue.KeyValueBackupTask.revertTask(KeyValueBackupTask.java:1239)
W BackupTransportClient: 	at com.android.server.backup.keyvalue.KeyValueBackupTask.run(KeyValueBackupTask.java:387)
W BackupTransportClient: 	at java.lang.Thread.run(Thread.java:1012)
I KeyValueBackupTask: K/V backup pass finished
I NotificationBackupObserver: Backup finished 1/[redacted total]. Status: -1000
```

grote:
Good find @t-m-w! I think this is the same issue we found when porting to Android 14: https://github.com/seedvault-app/seedvault/issues/564#issuecomment-1758222387

There's so many APKs to backup for this user and the connection is probably too slow to finish all of them within the transport timeout.

[...]

PR #628, opened 2024-02-22
Move to our own scheduling

### LINKED ISSUE #583: Move to our own scheduling
We can now **disable framework scheduling of backups** via  `IBackupManager#setFrameworkSchedulingEnabledForUser()` and currently have these use-cases lined up for using our own scheduling:
   * not having to hack into backup API methods for doing things like APK backups
   * not trying to back-up when location is on flash drive
   * not trying to back-up when location is on local server while in different network
   * doing our  own scheduling, so we can control `BackupDestination` without OS patches
   * doing our  own scheduling, so users can get more control about when backups are done
   
Storage backups are already using their own scheduling.

grote:
Closes #583, #629, #590


app/src/main/java/com/stevesoltys/seedvault/UsbIntentReceiver.kt
```diff
@@ -63,9 +63,7 @@ class UsbIntentReceiver : UsbMonitor() {
             i.putExtra(EXTRA_START_APP_BACKUP, true)
             startForegroundService(context, i)
         } else {
-            Thread {
-                requestBackup(context)
-            }.start()
+            AppBackupWorker.scheduleNow(context, reschedule = false)
```
stevesoltys, line 66: Maybe this should use the feature flag?

app/src/main/java/com/stevesoltys/seedvault/App.kt
```diff
@@ -115,6 +126,25 @@ open class App : Application() {
...
+     */
+    protected open fun migrateToOwnScheduling() {
+        if (!isFrameworkSchedulingEnabled()) return // already on own scheduling
+
+        backupManager.setFrameworkSchedulingEnabledForUser(UserHandle.myUserId(), false)
+        if (backupManager.isBackupEnabled) {
+            AppBackupWorker.schedule(applicationContext, settingsManager, UPDATE)
+        }
+        // cancel old D2D worker
+        WorkManager.getInstance(this).cancelUniqueWork("APP_BACKUP")
```
stevesoltys, line 141: Should this be a constant in `AppBackupWorker`?

app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
```diff
@@ -192,9 +174,60 @@ internal class MetadataManager(
...
 
+    /**
+     * Call this for all packages we can not back up for some reason.
+     *
+     * It updates the packages' local metadata.
+     * You still need to call [uploadMetadata] to persist all local modifications.
+     */
+    @Synchronized
+    @Throws(IOException::class)
+    internal fun onPackageDoesNotGetBackedUp(
```
stevesoltys, line 180: Maybe just `updatePackageState`? This indicates what it does in the function name, rather than where it's being called from.

app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
```diff
@@ -192,9 +174,60 @@ internal class MetadataManager(
...
+        packageInfo: PackageInfo,
+        packageState: PackageState,
+    ) = modifyCachedMetadata {
+        val packageName = packageInfo.packageName
```
stevesoltys:
```
        metadata.packageMetadataMap.getOrPut(packageInfo.packageName) {
            PackageMetadata(
                time = 0L,
                state = packageState,
                system = packageInfo.isSystemApp(),
            )
        }.state = packageState
```

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsActivity.kt
```diff
@@ -36,6 +37,19 @@ class SettingsActivity : RequireProvisioningActivity(), OnPreferenceStartFragmen
...
+
+        // observe initialization and show/remove init fragment
+        // this can happen when enabling backup and storage wasn't initialized
+        viewModel.initEvent.observeEvent(this) { show ->
+            val tag = "INIT"
+            if (show) {
+                val title = getString(R.string.storage_check_fragment_backup_title)
+                showFragment(StorageCheckFragment.newInstance(title), true, tag)
+            } else {
+                val f = supportFragmentManager.findFragmentByTag(tag)
```
stevesoltys: Maybe `fragment` as variable name, as single letter variable names can be considered confusing?

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsActivity.kt
```diff
@@ -36,6 +37,19 @@ class SettingsActivity : RequireProvisioningActivity(), OnPreferenceStartFragmen
...
+                if (f != null && f.isVisible) supportFragmentManager.popBackStack()
```
stevesoltys: Could be `if(fragment?.isVisible == true)`.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsFragment.kt
```diff
@@ -244,10 +255,41 @@ class SettingsFragment : PreferenceFragmentCompat() {
...
-        backupStatus.summary = getString(R.string.settings_backup_status_summary, lastBackup)
+    private fun setAppBackupStatusSummary(lastBackupInMillis: Long?) {
+        if (lastBackupInMillis != null) {
+            // set time of last backup
+            val lastBackup = lastBackupInMillis.toRelativeTime(requireContext())
+            backupStatus.summary = getString(R.string.settings_backup_status_summary, lastBackup)
+        }
+    }
+
+    private fun setAppBackupSchedulingSummary(workInfo: WorkInfo?) {
```
stevesoltys, line 273: This function is on the larger side, maybe could use some documentation?

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
```diff
@@ -134,13 +146,24 @@ class SettingsManager(private val context: Context) {
...
-        return !storage.isUnavailableUsb(systemContext) && !storage.isUnavailableNetwork(context)
+        return !storage.isUnavailableUsb(systemContext) &&
+            !storage.isUnavailableNetwork(context, useMeteredNetwork)
     }
 
     fun backupApks(): Boolean {
         return prefs.getBoolean(PREF_KEY_BACKUP_APK, true)
     }
 
+    val backupFrequencyInMillis: Long
```
stevesoltys, line 157: I assume we plan on exposing this to the user?

app/src/main/java/com/stevesoltys/seedvault/settings/ExpertSettingsFragment.kt
```diff
@@ -16,10 +16,10 @@ class ExpertSettingsFragment : PreferenceFragmentCompat() {
     private val viewModel: SettingsViewModel by sharedViewModel()
     private val packageService: PackageService by inject()
 
-    // TODO set mimeType when upgrading androidx lib
-    private val createFileLauncher = registerForActivityResult(CreateDocument()) { uri ->
-        viewModel.onLogcatUriReceived(uri)
-    }
+    private val createFileLauncher =
+        registerForActivityResult(CreateDocument("text/plain")) { uri ->
```
stevesoltys: `org.apache.http.entity.TEXT_PLAIN.mimeType` maybe or something similar

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsFragment.kt
```diff
@@ -244,10 +255,41 @@ class SettingsFragment : PreferenceFragmentCompat() {
...
+        if (storage?.isUsb == true) {
+            backupScheduling.summary = getString(R.string.settings_backup_status_next_backup_usb)
+            return
+        }
+        if (workInfo == null) return
```
stevesoltys, line 278: Do we need to clear any existing state?

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
```diff
@@ -120,18 +116,15 @@ internal class BackupCoordinator(
...
         // instead of simply deleting the current one
-        val token = startNewRestoreSet()
+        startNewRestoreSet()
         Log.i(TAG, "Initialize Device!")
         plugin.initializeDevice()
-        Log.d(TAG, "Resetting backup metadata for token $token...")
-        plugin.getMetadataOutputStream(token).use {
-            metadataManager.onDeviceInitialization(token, it)
-        }
+        startNewRestoreSet()
```
stevesoltys: We are starting two restore sets, intentionally I suppose? Maybe a comment on why?

stevesoltys (dismissed):
Mainly just nitpicking.

app/src/main/java/com/stevesoltys/seedvault/UsbIntentReceiver.kt
grote: Why a feature flag? This is the same behavior as before. When the user plugs in the flash drive and sufficient time has passed, it requests a backup *now*.

app/src/main/java/com/stevesoltys/seedvault/App.kt
grote: This is migratory code, so I had decided against it. This is the only time we'll ever going to use it. This is for people that had tried the d2d feature flag and used a worker before. This is now superseded by the new workers.

The real unique work IDs we are still using have constants.

app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
grote: I am a bit torn here, as all the other methods are also using names related to the purpose of the call rather the implementation (e.g. `onPackageBackupError()`).

app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
grote: good call. I used getOrPut() here and in the other places where it made sense.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsActivity.kt
grote: Sure!

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsActivity.kt
grote: Not sure which is more readable, but ok.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsFragment.kt
grote: Alright, added.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
grote: It is already exposed and configurable via `SchedulingFragment`.

app/src/main/java/com/stevesoltys/seedvault/settings/ExpertSettingsFragment.kt
grote: I found `com.google.android.mms.ContentType.TEXT_PLAIN` and now use that, but not sure if that's better than just leaving the string in.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsFragment.kt
grote: Good question. I don't *think* we need to, because it should only ever be `null` when the workInfo wasn't loaded. Afterwards, we *should* always get something non-null here.

app/src/main/java/com/stevesoltys/seedvault/transport/backup/BackupCoordinator.kt
grote: No, I think this is an oversight on my part. I'll take a look at that separately, as this is more delicate.

grote: Yeah, so this wasn't correct. Well spotted. Now, we are only doing this once. I tested it and it should work.

theimpulson (approved):
Looks good to me. Feel free to merge when ready.

app/src/main/java/com/stevesoltys/seedvault/App.kt
theimpulson: Marking as resolved assuming we are fine with this here.

PR #632, opened 2024-03-08
Fix flakyness in SmallFileBackupIntegrationTest

grote:


storage/lib/src/test/java/org/calyxos/backup/storage/backup/SmallFileBackupIntegrationTest.kt
```diff
@@ -93,7 +92,14 @@ internal class SmallFileBackupIntegrationTest {
         every { mac.doFinal(any<ByteArray>()) } returns chunkId
         every { chunksCache.get(any()) } returns null
         every { storagePlugin.getChunkOutputStream(any()) } returns outputStream2
-        every { chunksCache.insert(cachedChunk) } just Runs
+        every {
+            chunksCache.insert(match<CachedChunk> { cachedChunk ->
+                cachedChunk.id == chunkId.toHexString() &&
+                    cachedChunk.refCount == 0L &&
+                    cachedChunk.size <= outputStream2.size() &&
```
grote, line 99: since the chunk gets zipped, the random input data would sometimes differ in size if randomness wasn't truly random, so it could be compressed more.

chirayudesai (approved):

PR #646, opened 2024-04-16
Fix nullability for UserManager in AOSP build

grote:


app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
```diff
@@ -312,7 +312,7 @@ internal class MetadataManager(
     private fun getUserName(): String? {
         val perm = "android.permission.QUERY_USERS"
         return if (context.checkSelfPermission(perm) == PERMISSION_GRANTED) {
-            val userManager = context.getSystemService(UserManager::class.java)
+            val userManager = context.getSystemService(UserManager::class.java) ?: return false
```
t-m-w: Am I misunderstanding this, or is this trying to return `false` as a `String?`?

t-m-w (changes requested):

app/src/main/java/com/stevesoltys/seedvault/metadata/MetadataManager.kt
grote: that was supposed to be `null`. sorry. I did this right from the web ui as I am working on another branch at the moment.

theimpulson (approved):
Looks good to me

t-m-w:
Might be good to remove the first commit from this? Or will they get squashed automatically?

chirayudesai:
> Might be good to remove the first commit from this? Or will they get squashed automatically?

We can squash from the UI when merging, which I did.

PR #647, opened 2024-04-16
Implement a native WebDAV plugin

### LINKED ISSUE #317: Implement a generic WebDAV backend
* [x] Refactor Seedvault’s storage and UI code to allow for different (non-SAF) storage plugins
  * [x]  don't allow changing storage while file or app backup is running
* [x] Implement new WebDAV storage plugin
* [x] UI for adding WebDAV credentials and URL

---

Hey,

first of all thanks for providing a FOSS backup solution for Android! Much appreciated!

As already mentioned in #84 I would like to request a generic WebDAV backend as storage provider.

I think this would come in handy to many users since it enables to use any cloud solution as backup target as long it supports WebDAV (many FOSS cloud solutions like Seafile, OwnCloud, NextCloud, Syncthing support WebDAV).

When selecting the WebDAV backend as storage provider a small configuration panel should open in which server, port, https (yes/no), path on the server, username and password can be entered. 

Thanks!

grote:
This MR does the following:

* Refactor Seedvault’s storage and UI code to allow for different (non-SAF) storage plugins
  *    don't allow changing storage while file or app backup is running
* Implement new WebDAV storage plugin
* UI for adding WebDAV credentials and URL

Closes #317


app/src/main/java/com/stevesoltys/seedvault/plugins/StoragePlugin.kt
```diff
@@ -1,13 +1,17 @@
...
 import java.io.OutputStream
 
-interface StoragePlugin {
+interface StoragePlugin<T> {
+
+    /**
+     * Returns true if the plugin is working, or false if it isn't.
+     * @throws Exception any kind of exception to provide more info on the error
+     */
+    suspend fun test(): Boolean
```
stevesoltys, line 14: Wonder if something like [Result4k](https://github.com/fork-handles/forkhandles/tree/trunk/result4k) would be helpful here?

app/src/main/java/com/stevesoltys/seedvault/plugins/StoragePluginManager.kt
```diff
@@ -0,0 +1,162 @@
...
+import com.stevesoltys.seedvault.getStorageContext
+import com.stevesoltys.seedvault.permitDiskReads
+import com.stevesoltys.seedvault.plugins.saf.DocumentsProviderStoragePlugin
+import com.stevesoltys.seedvault.plugins.saf.DocumentsStorage
+import com.stevesoltys.seedvault.plugins.saf.SafFactory
+import com.stevesoltys.seedvault.plugins.webdav.WebDavFactory
+import com.stevesoltys.seedvault.settings.SettingsManager
+import com.stevesoltys.seedvault.settings.StoragePluginEnum
+
+abstract class StorageProperties<T> {
```
stevesoltys: Maybe should go in its own file?

app/src/main/java/com/stevesoltys/seedvault/plugins/StoragePluginManager.kt
```diff
@@ -0,0 +1,162 @@
...
+}
+
+class StoragePluginManager(
+    private val context: Context,
+    private val settingsManager: SettingsManager,
+    safFactory: SafFactory,
+    webDavFactory: WebDavFactory,
+) {
+
+    private var _appPlugin: StoragePlugin<*>?
```
stevesoltys: Do we plan on prefixing with `_` for private variables throughout the app? I think we don't do that elsewhere right now do we?

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
```diff
@@ -22,6 +24,13 @@ internal const val PREF_KEY_SCHED_FREQ = "scheduling_frequency"
 internal const val PREF_KEY_SCHED_METERED = "scheduling_metered"
 internal const val PREF_KEY_SCHED_CHARGING = "scheduling_charging"
 
+private const val PREF_KEY_STORAGE_PLUGIN = "storagePlugin"
+
+internal enum class StoragePluginEnum { // don't rename, will break existing installs
```
stevesoltys: `StoragePluginType` maybe?

app/src/main/java/com/stevesoltys/seedvault/plugins/webdav/WebDavStorage.kt
```diff
@@ -0,0 +1,181 @@
...
+    protected val okHttpClient = OkHttpClient.Builder()
+        .followRedirects(false)
+        .authenticator(authHandler)
+        .addNetworkInterceptor(authHandler)
+        .connectTimeout(15, TimeUnit.SECONDS)
+        .writeTimeout(30, TimeUnit.SECONDS)
+        .readTimeout(120, TimeUnit.SECONDS)
+        .pingInterval(45, TimeUnit.SECONDS)
+        .connectionSpecs(listOf(ConnectionSpec.MODERN_TLS))
+        .retryOnConnectionFailure(true)
```
stevesoltys, line 62: Do we need to specify a retry limit? Is the default sufficient?

app/src/main/java/com/stevesoltys/seedvault/plugins/webdav/WebDavStoragePlugin.kt
```diff
@@ -0,0 +1,199 @@
...
+) : WebDavStorage(webDavConfig, root), StoragePlugin<WebDavConfig> {
+
+    override suspend fun test(): Boolean {
+        val location = baseUrl.toHttpUrl()
+        val davCollection = DavCollection(okHttpClient, location)
+
+        val webDavSupported = suspendCoroutine { cont ->
+            davCollection.options { davCapabilities, response ->
+                debugLog { "test() = $davCapabilities $response" }
+                if (davCapabilities.contains("1")) cont.resume(true)
```
stevesoltys, line 37: What do these checks do? Couldn't we do `containsAny` or similar?

app/src/main/java/com/stevesoltys/seedvault/storage/WebDavStoragePlugin.kt
```diff
@@ -0,0 +1,290 @@
...
+            getOutputStream(location)
+        } catch (e: Exception) {
+            if (e is IOException) throw e
+            else throw IOException("Error getting OutputStream for $chunkId: ", e)
+        }
+    }
+
+    @Throws(IOException::class)
+    override suspend fun getBackupSnapshotOutputStream(timestamp: Long): OutputStream {
+        val location = "$url/$folder/$timestamp.SeedSnap".toHttpUrl()
```
stevesoltys: Are we using the file extension `SeedSnap`? Do we need this to be a constant to avoid mistakes?

stevesoltys (dismissed):

app/src/main/java/com/stevesoltys/seedvault/plugins/StoragePlugin.kt
grote: yeah not only here, but all over the codebase. so far it is mostly following the old java paradigm, that you have to catch and handle exceptions everywhere (except the compile isn't insisting on it).

app/src/main/java/com/stevesoltys/seedvault/plugins/StoragePluginManager.kt
grote: Done in extra commit.

app/src/main/java/com/stevesoltys/seedvault/plugins/StoragePluginManager.kt
grote: This is not about private variables, this is for fields that have a non-mutable public companion. We do this elsewhere, but it seems the `mFoo` naming is more prevalent, so I added a commit that renames the `_foo` occurrences (although I like them better).

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsManager.kt
grote: good call. renamed.

app/src/main/java/com/stevesoltys/seedvault/plugins/webdav/WebDavStorage.kt
grote: so far I've found the default to work nicely. But we can tweak this as the need arises. However, the okhttp authors are probably better at tweaking this than me.

app/src/main/java/com/stevesoltys/seedvault/plugins/webdav/WebDavStoragePlugin.kt
grote: This checks, if the storage supports webdav. I think the `1` refers to the DAV Compliance Class 1 in https://www.ietf.org/rfc/rfc4918.txt

`containsAny` would work as well, but where do you get that from?

In addition: Users that enter the wrong URL for nextcloud, e.g. `example.org` instead of `example.org/remote.php/webdav`, this test will pass but later the storage won't be usable, because we can't create a folder there. So I wonder if creation of our root folder should be part of the test.

app/src/main/java/com/stevesoltys/seedvault/storage/WebDavStoragePlugin.kt
grote: Alright, pushed a commit that does this project-wide.

app/src/main/java/com/stevesoltys/seedvault/plugins/webdav/WebDavStoragePlugin.kt
stevesoltys: Got it thanks. I was just a bit confused as someone reading the code as to what this does, so I figure it could it be helpful to provide the above information in a comment or maybe make these constants to indicate what they are.

stevesoltys (approved):

stevesoltys:
Feel free to merge whenever. LGTM.

app/src/test/java/com/stevesoltys/seedvault/plugins/webdav/WebDavTestConfig.kt
```diff
@@ -0,0 +1,22 @@
...
+package com.stevesoltys.seedvault.plugins.webdav
+
+import org.junit.Assume.assumeFalse
+import org.junit.jupiter.api.Assertions.fail
+
+object WebDavTestConfig {
+
+    fun getConfig(): WebDavConfig {
+        assumeFalse(System.getenv("NEXTCLOUD_URL").isNullOrEmpty())
+        return WebDavConfig(
```
stevesoltys, line 15: Another option could be something like WireMock if we didn't want to actually depend on an external service for tests.

But this is probably easier and likely provides better coverage anyways..

grote: Yeah, I had considered mocking the HTTP interactions, but this gets us to low-level DAV and XML details, I didn't want to get into. That is why we are using a DAV library after all. I also thought about some built-in WebDAV capable webserver only used for tests, but couldn't find anything sane and easy, so I went with testing against a real server.

PR #657, opened 2024-05-07
Add helper text for WebDAV URL field

grote:
Screenshot_20240507_160201


t-m-w:
This already worked for me! "Worked" as in, even though I had pulled this change, I still forgot to put the full URL for Nextcloud. After it showed the error, a combination of memory + seeing this message helped me to fix the URL, and I didn't have to look it up.

chirayudesai (approved):

chirayudesai:
Since the URL is only meant to serve as an example and not meant to be clickable, this should be fine.

PR #663, opened 2024-05-18
Make seedvault almost REUSE compliant

mikeNG:


mikeNG:
I've had these commits locally for over a year now, I rebased them on top of the current HEAD and fixed conflicts and added all the new files in the 4/5 and 5/5 commits. The copyright dates have been added according to file creation date.

At the moment the only remaining issue is these two files have no license specified:
* app/libs/android.jar
* app/libs/libcore.jar

However it's pretty difficult to define a license for them especially due to the complicated copyright of libcore.

Any comments, suggestions or edits are welcome

grote:
> However it's pretty difficult to define a license for them especially due to the complicated copyright of libcore.

You probably know the situation there better than me.

Is there an issue with leaving those two out for the time being?

chirayudesai (approved):

chirayudesai:
> At the moment the only remaining issue is these two files have no license specified:

We can figure that out but I don't want to block this any further, let's merge it, we can always do another PR right after to address the remaining.

Since this is just about compliance anyways it doesn't hurt having something partial in at this point.

Issue #624, opened 2024-02-13
Show system apps in Backup status


chirayudesai:
At the moment, we don't show them at all.

We can't show all system apps since there's too many, and a lot of them don't even make sense to show (overlays, etc)

For a start, having just system apps with icons shown as a separate list at the bottom could work.

chirayudesai:
Test: Especially after D2D, Chromium and friends / forks are able to get backed up and restored.

grote:
When you say with icons, do you mean show/count all system apps with *launcher* icons?

chirayudesai:
> system apps with *launcher* icons?

Yes

The idea being those are the apps easily visible

grote:
Any idea how to identify those? Do we need to query all intent activities and then cache the list of packages with a launcher activity or given we have `PackageInfo` already, is there a way to find out if the package has a launcher activity?

chirayudesai:
> Any idea how to identify those? 

`android.intent.category.LAUNCHER` would be my first pick.

I was thinking of the Firewall at first but that simply lists all system apps, which would lead to a long pointless list here.

Maybe another way is all system apps which have some data to back up, if that's easier. That then easily excludes a good chunk.

dot166:
> > Any idea how to identify those?
> 
> `android.intent.category.LAUNCHER` would be my first pick.
> 
> I was thinking of the Firewall at first but that simply lists all system apps, which would lead to a long pointless list here.
> 
> Maybe another way is all system apps which have some data to back up, if that's easier. That then easily excludes a good chunk.

What about Launcher3? It (obviously) does not have an intent that has the `android.intent.category.LAUNCHER` catagory because it is the launcher. Also there is data that can be backed up from it (layout, wallpaper, icon pack (if applicable), etc).

grote:
since https://github.com/seedvault-app/seedvault/pull/670, we now show (launchable) system apps in backup status as requested.

PR #670, opened 2024-05-30
Allow choosing which apps will get restored

### LINKED ISSUE #309: Choose what apps will get restored
* [ ] New restore UI allowing selection of apps to restore before actual restore process starts
* [ ] Enable restore in normal UI, so users can restore (non-system?) apps even after completing their device setup wizard

---

Somewhat related to https://github.com/seedvault-app/seedvault/issues/100

When I restore a backup, maybe I don't want to restore some apps, I used to have before. IIRC this option was not offered but I know some device transfer services/apps like from Samsung e.g. offer this.

The reasons/use cases may be:
* **technical:** I know the app is only compatible with my old device/StockROM/CustomROM and got accidentally backed up (or in the future because I do things like https://github.com/seedvault-app/seedvault/issues/165)
* **aesthetically:** I want to clean up my system and use this time to start with a smaller step of apps.
  IMHO backup restoration is a good point to reconsider ones decisions about how useful an app was/is/how often it had been used and whether you still need it. Just some chore…
  To support that feature, of course it would be awesome if there was "App was last used three days/months/years ago" or so, if the app was not used for a longer time. But that's rather off-topic…
  BTW for the backing up step of the apps I would not consider this relevant as I rather think that a backup should rather include more data than too less. (that said, I know, of course you can actually already exclude apps/things from being restored here in Seedvault)
* **practically:** I may just want to install some apps, because I'm an CustomROM developer/tester or so and always have my same backup and 5 different devices/emulators or so. For that, I just want a specific set of apps to install (without having to do that manually).
  Or I may have any other obscure reason not to restore a full backup.
  
So it could just present a list of all things pre-selected and let me disable some, I don't want to restore…
I quite wonder why this feature has not been proposed yet, at least I did not find anything in the issue tracker.

grote:
This adds quite a bit of plumping around the restore process. E.g. we now separately store all the icons of apps we back up, as well as their names and system app status. Apps during restore are now also (mostly) sorted by name.

This then leads to a bit of changes in how we display backup status, e.g. check-boxes instead of switches, show system data instead of system apps and include launch-able system apps in list.

### Testing

* check that backup status list looks as expected
* check that apps can still be excluded from backup
* test restore process in various scenarios
  * from existing installation
  * from new install SUW
  * with system app restore and without (while restoring some shown system apps and others not)
  * check that unselected apps don't get restored
  * with APK backup enabled and without
  * check that APKs get installed fine
  * check that APKs that could not get installed or were not backed up can manually be installed
  * check that a backup made with an older version restores just as fine as one made with the current version

Closes #309, #624


t-m-w:
> with system app restore and without (while restoring some shown system apps and others not)

When restoring, I see a "System apps" toggle under "System data", but I do not see the ability to select or deselect individual system apps. What I'm looking for is e.g. Chromium, but I don't see it. Am I missing something?

Edit: I do see some, such as Phone and Music, mixed in with other user apps (that's the part I wasn't expecting, but it's fine). It turns out that Calculator, Chromium, Clock, and some others are "Waiting to back up..." for some reason. I do have D2D on, currently.

t-m-w:
UX: As I scroll around the list of apps to restore, when I scroll down, including when I reach the bottom, there is no clear way to proceed with restoration, as the "Restore backup" button hides itself when you scroll down. I need to scroll up a bit to make the floating "Restore backup" button reappear, which is a bit unintuitive.

grote:
> When restoring, I see a "System apps" toggle under "System data", but I do not see the ability to select or deselect individual system apps. What I'm looking for is e.g. Chromium, but I don't see it. Am I missing something?
> 
> Edit: I do see some, such as Phone and Music, mixed in with other user apps (that's the part I wasn't expecting, but it's fine). It turns out that Calculator, Chromium, Clock, and some others are "Waiting to back up..." for some reason. I do have D2D on, currently.

We do show launchable system apps like regular apps now. Other non-userfacing system apps are behind that "System apps" toggle as an all or nothing.

> UX: As I scroll around the list of apps to restore, when I scroll down, including when I reach the bottom, there is no clear way to proceed with restoration, as the "Restore backup" button hides itself when you scroll down. I need to scroll up a bit to make the floating "Restore backup" button reappear, which is a bit unintuitive.

Maybe @theimpulson knows a coordinator layout behavior that works better here. Because of the long app list and the need to go through all of it, I didn't want to steal precious screen space for that button and hence let it float above the list.

t-m-w:
When trying to restore an older backup (from CalyxOS 5.7.1, Pixel 8a akita), I do not see app names or icons, only package names and generic icons. Waiting a while does not cause them to load.

grote:
> When trying to restore an older backup (from CalyxOS 5.7.1, Pixel 8a akita), I do not see app names or icons, only package names and generic icons. 

Yes that is expected. Only the new code does store names and icons. As we don't control old installs and old backups, there's no way we can retroactively store those things. So restoring old backups will need to live without icons and app names.

Both do exist in the APKs, if available, but downloading all APKs before the user can select something is a bit much, so we didn't do it.

t-m-w:
Okay, everything checks out to me except for one thing which isn't really a Seedvault problem / not related to this change. I left a bullet next to it.

* [x] check that backup status list looks as expected
* [x] check that apps can still be excluded from backup
* test restore process in various scenarios
  * [x] from existing installation
  * [x] from new install SUW
  * [x] with system app restore and without (while restoring some shown system apps and others not)
  * [x] check that unselected apps don't get restored
  * [x] with APK backup enabled and without
  * [x] check that APKs get installed fine
  * [ ] check that APKs that could not get installed or were not backed up can manually be installed
    * [CalyxOS-specific failure](https://gitlab.com/CalyxOS/calyxos/-/issues/2437): F-Droid Basic and Aurora Store are not available at this time. Seedvault launches Chromium to a Play Store page, where nothing can be done.
    * If you do an existing-install restore with `*#*#RESTORE#*#*`, it's okay, except that it immediately fails for apps which have no APKs backed up, even if those apps were installed in the interim - so you tap to install, go back without doing anything, and the app moves to the bottom with a green checkmark. Unrelated issue I think, I believe I've seen it before.
  * [x] check that a backup made with an older version restores just as fine as one made with the current version
    * System apps are not listed individually, but this is likely not possible or feasible.

grote:
Thanks a lot for testing @t-m-w!

> If you do an existing-install restore with *#*#RESTORE#*#*, it's okay, except that it immediately fails for apps which have no APKs backed up, even if those apps were installed in the interim - so you tap to install, go back without doing anything, and the app moves to the bottom with a green checkmark. Unrelated issue I think, I believe I've seen it before.

Right, that's unrelated and one of the items of https://github.com/seedvault-app/seedvault/issues/671

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/SeedvaultLargeTest.kt
```diff
@@ -44,6 +45,11 @@ internal abstract class SeedvaultLargeTest :
         resetApplicationState()
         clearTestBackups()
 
+        runCommand("bmgr enable true")
+        sleep(60_000)
```
stevesoltys, line 49: Fine for now but this could become some sort of logcat grep... maybe possible...

app/src/main/java/com/stevesoltys/seedvault/ui/SystemData.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
+import com.stevesoltys.seedvault.R
+
+internal const val PACKAGE_NAME_SMS = "com.android.providers.telephony"
+internal const val PACKAGE_NAME_SETTINGS = "com.android.providers.settings"
+internal const val PACKAGE_NAME_CALL_LOG = "com.android.calllogbackup"
+internal const val PACKAGE_NAME_CONTACTS = "org.calyxos.backup.contacts"
+internal const val PACKAGE_NAME_SYSTEM = "@org.calyxos.system@"
```
stevesoltys, line 16: Is this specific to CalyxOS?

stevesoltys (approved):

app/src/androidTest/java/com/stevesoltys/seedvault/e2e/SeedvaultLargeTest.kt
grote: yeah would be nice to shorten this (and other) times.

app/src/main/java/com/stevesoltys/seedvault/ui/SystemData.kt
grote: No, it is just a fake package name like `@pm@` and just matters at runtime, I had `@system@` first but then was afraid Android ever starts using it for something. We could use `@org.stevesoltys.seedvault.system@` as well.

PR #672, opened 2024-05-31
APK/App Backups: Improve wording

lucasmz-dev:
This pull request attempts at improving wording when it comes to APK backups.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -26,12 +26,12 @@
     <string name="settings_auto_restore_summary">When reinstalling an app, restore backed up settings and data.</string>
     <string name="settings_auto_restore_summary_usb">Note: Your %1$s needs to be plugged in for this to work.</string>
     <string name="settings_category_app_data_backup">App data backup</string>
-    <string name="settings_backup_apk_title">App backup</string>
+    <string name="settings_backup_apk_title">APK backup</string>
     <string name="settings_backup_apk_summary">Back up the apps themselves. Otherwise, only app data would get backed up.</string>
-    <string name="settings_backup_apk_dialog_title">Really disable app backup?</string>
-    <string name="settings_backup_apk_dialog_message">Disabled app backup will still back up app data. However, it will not get restored automatically.\n\nYou will need to install all your apps manually while having \"Automatic Restore\" switched on.</string>
+    <string name="settings_backup_apk_dialog_title">Really disable APK backup?</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
lucasmz-dev, line 1: APK backups might not be a great name, it'll be nice to use the older scheme but in a better fashion, I'm not sure how to go about that though.

lucasmz-dev:
Right now, there are two strings which are the same, "App backup" one used for separating settings, and one that's for APK backups.
One thing that this definitely caused confusions, was with translations, at least for me.

Currently, this PR only changes the strings to replace 'app' with 'APKs'. 

I ended up using 'APK' wording because I feel like most people that are gonna end up using Seedvault do know what an APK is, they've probably gone out of the way of a custom ROM, and regardless, there's also the text explaining that that option is used for backing the apps themselves, which is currently the only thing that really explains what this option does properly.

grote:
We intentionally avoided the term APK, because we think that most people (normal users) don't necessarily know what that is. So we made the distinction between backup the apps themselves and their data. This could be made clearer in our strings and maybe we can include (APKs) somewhere in brackets, but we shouldn't rely on people knowing what that means.

lucasmz-dev:
I was looking into other ways of making a better distinction, and the other way I could think about is to specify when we mean 'app data'; the problem is that Seedvault works in a way where it's not like you can backup just the APKs, so people might get confused there. It's a sub-config of app backups, not a separate thing;

I think that ultimately it might just be better to use APK (or something like app package?) wording, we wouldn't have to rely on it, as the description explains well what the option does and so it is quite easy to infer what an APK would be; people really only have to know that this option means that the apps themselves get backed up.

All strings right now seem to be just things that show up right away after clicking that option.

At the moment, I find the wording from APK and app data backups to collide way too much, from the category being called 'App backup' and also the APK backup option, sub-menus using the same terms, except with a difference in plural use; I think ultimately using APK or package or whatever, has a better effect than just keeping it as 'App backup' and having its wording collide constantly with app data backups.

The current wording causes more confusion than if people were confused on what an APK is IMO.

https://github.com/seedvault-app/seedvault/assets/55422065/8f7e2860-6311-494e-b3db-016d7fba8fd2

I marked this as a draft before to see if I could get some sort of alternative, but I think this might just be the best way to do this, as it doesn't mix settings together. So I'll just leave this open now with the notes

---

I also did this inside the PT-BR translation (which I feel like was more complicated to distinguish these), and I found it much more clear. I undid it in Weblate to see how it works here, in the meanwhile.

PR #691, opened 2024-06-18
Show full URL in settings when WebDAV is used

### LINKED ISSUE #690: Show full backup location, particularly for WebDAV
As I've been testing things, I have a device which is set to use WebDAV and a particular server, but I cannot remember the full backup location it's set to use, and I cannot view it. All I can see in the Seedvault settings preference for Backup location is "WebDAV [hostname]". Being able to see the full path would be useful for me in this case.

(Also being able to see the username would be useful, but that's a bonus.)

grote:
Showing just the host is good for most people, I'd say and showing the full path may be ugly and un-needed in case of Nextcloud, so not sure this change is good.

Fixes #690

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/backend/webdav/WebDavHandler.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/backend/webdav/WebDavHandler.kt
@@ -6,7 +6,6 @@
 package com.stevesoltys.seedvault.backend.webdav
 
 import android.content.Context
-import android.net.Uri
 import android.util.Log
 import androidx.annotation.WorkerThread
 import com.stevesoltys.seedvault.R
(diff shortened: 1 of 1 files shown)


chirayudesai:
> showing the full path may be ugly and un-needed

I had that thought at first too, but then I'd say it'd be better to put function over form, show this now, and figure out a better way to tweak this change later if wanted.

chirayudesai (dismissed):

t-m-w (dismissed):

chirayudesai:
Bump!

t-m-w (dismissed):
I've been using this (except when I forgot to pull it) and it works fine. I think it's worthwhile.

grote:
I resolved a conflict. @t-m-w does this still work as expected?

t-m-w (approved):
This should be fine. The way I "resolved" this conflict on my side last time was to just do `val host = config.url`, not even remove the prefix, since it doesn't bother me, and that was fine. So I don't see why this wouldn't be fine, too.

PR #702, opened 2024-06-27
Improve logging relevant for WebDAV streams

grote:
When writing to WebDAV streams, we usually only get a PipeClosed error without information what *actually* went wrong. This PR aims to improve this situation.


t-m-w (approved):
Works for me!

PR #705, opened 2024-07-03
 Allow choosing what files/folders will get restored

### LINKED ISSUE #565: Choose what files/folders will get restored
When restoring backed up files, Seedvault currently only offers an all or nothing UI. We want to improve this, so users can choose which files they want to restore.

* [ ] New restore UI allowing selection of files/folders to restore before actual restore process starts
* [ ] Enable restore in normal UI, so users can restore files even after completing their device setup wizard. This needs to warn about overwriting existing files or creating duplicates

grote:
https://github.com/seedvault-app/seedvault/assets/244947/dba1b255-f0db-4834-a825-f82429b43eec

Fixes #565

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
@@ -224,12 +224,13 @@ internal class SettingsViewModel(
 
     internal fun backupNow() {
         viewModelScope.launch(Dispatchers.IO) {
+            val isAppBackupEnabled = backupManager.isBackupEnabled
             if (settingsManager.isStorageBackupEnabled()) {
                 val i = Intent(app, StorageBackupService::class.java)
-                // this starts an app backup afterwards
-                i.putExtra(EXTRA_START_APP_BACKUP, true)
+                // this starts an app backup afterwards (if enabled)
+                i.putExtra(EXTRA_START_APP_BACKUP, isAppBackupEnabled)
                 startForegroundService(app, i)
-            } else {
+            } else if (isAppBackupEnabled) {
                 AppBackupWorker.scheduleNow(app, reschedule = !pluginManager.isOnRemovableDrive)
             }
         }
--- a/storage/lib/src/main/java/org/calyxos/backup/storage/ui/restore/FileSelectionManager.kt
+++ b/storage/lib/src/main/java/org/calyxos/backup/storage/ui/restore/FileSelectionManager.kt
@@ -54,6 +54,7 @@ public class FileSelectionManager {
 
     private val allFolders = HashMap<String, FolderItem>()
     private val allFiles = HashMap<String, MutableList<FileItem>>()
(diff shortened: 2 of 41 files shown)


t-m-w (dismissed):
Tested the functionality of this change, and it looks good.

All selected files/folders were restored. Others were not restored. Functionality was as expected.

Everything noted is fairly minor. Most have already been fixed.

## Restore page
- Download directory shows a plain file icon for .jpg and .png files. Folders like DCIM/Camera show image icons.

### UX
These were fixed already:
- [x] Tapping a file does nothing, but tapping a directory expands or collapses it. It may make sense for tapping the file to change its selected state.
- [x] If there are no files to restore, the page is just empty, including the typical "Restore selected files" files button, etc., but nothing to select.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
```diff
@@ -224,12 +224,13 @@ internal class SettingsViewModel(
...
+            val isAppBackupEnabled = backupManager.isBackupEnabled
             if (settingsManager.isStorageBackupEnabled()) {
                 val i = Intent(app, StorageBackupService::class.java)
-                // this starts an app backup afterwards
-                i.putExtra(EXTRA_START_APP_BACKUP, true)
+                // this starts an app backup afterwards (if enabled)
+                i.putExtra(EXTRA_START_APP_BACKUP, isAppBackupEnabled)
                 startForegroundService(app, i)
-            } else {
+            } else if (isAppBackupEnabled) {
```
stevesoltys, line 233: Seems unrelated to this PR sorta? But seems valid nonetheless. Is this a bugfix maybe?

stevesoltys (dismissed):

storage/lib/src/main/java/org/calyxos/backup/storage/ui/restore/FileSelectionManager.kt
```diff
@@ -0,0 +1,255 @@
...
+    val expanded: Boolean,
+) : FilesItem {
+    init {
+        check(selected || !partiallySelected) {
+            "$dir was not selected, but partially selected"
+        }
+    }
+}
+
+public class FileSelectionManager {
```
stevesoltys, line 15: Preference here, but since this is a larger class, it might benefit visually being separated from the object models above.

app/src/main/java/com/stevesoltys/seedvault/settings/SettingsViewModel.kt
grote: Yeah, I think I had noticed during testing and fixed along the way instead of PRing separately.

storage/lib/src/main/java/org/calyxos/backup/storage/ui/restore/FileSelectionManager.kt
grote: Alright, split it out into its own file.

stevesoltys (approved):

PR #712, opened 2024-07-31
Add launcher icon for debug builds

grote:
Note that release builds still won't have a launcher icon, because the functionality is embedded into the Settings app.


theimpulson:
@grote is it possible for you to also add a Android Studio task to launch the debug builds on device like normal app instead of launching emulator?

grote:
> @grote is it possible for you to also add a Android Studio task to launch the debug builds on device like normal app instead of launching emulator?

Not sure I understand. I have an `app` run configuration that launches on normal device, if connected. I am not even using emulators myself.

t-m-w (approved):
Works! `./gradlew installDebug` or `adb install` directed at the path work fine for the Seedvault app itself on userdebug build (unrelated contactsbackup signing config issues notwithstanding).

PR #715, opened 2024-08-02
Add design document for de-dup research

grote:


theimpulson:
Looks good to me

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+It is inspired by the design of the [Restic backup program](https://restic.net/).
+Portions of this document are blatant copies of
+[its documentation](https://restic.readthedocs.io/en/latest/100_references.html).
+However, due to the different nature of Android backups,
+the design [was heavily simplified](#differences-to-restic).
+
+Note that the backup of files is described [in a different document](../storage/doc/design.md).
+
+## Terminology
```
t-m-w, line 14: [Clarity] There are parts of the Terminology section that seem to shy away from using the word "file" when referring to backup storage. Is this intentional? Is it meant to be more generic so as to potentially apply to e.g. block storage backends as well?

If so, I understand, but if that's the case, the document has numerous other references to "files" that should possibly be altered to avoid confusion.

I've added comments in a few other, specific places related to this.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+This section introduces terminology used in this document.
+
+*Repository*: All data produced during a backup is sent to and stored in a repository
+in a structured form.
+
+*Chunk*: Larger files are cut into re-usable chunks that are the unit of de-duplication.
+
+*Blob*: A blob is a chunk stored encrypted and compressed in the repository.
```
t-m-w: [Clarity] A blob is a file, right? Maybe clarify: "A blob is a chunk stored compressed and encrypted as an individual file in the repository."

This clarifies that a blob is a file, and (nit) that compression happens before encryption, as it presumably does.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+*Snapshot*: A snapshot stands for the state of a collection of apps
```
t-m-w, line 25: [Clarity] If a blob is a file, and a snapshot is a file, then the definition of a snapshot ought to similarly emphasize that it is encrypted.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+that have been backed up at some point in time.
+The state here means the app data as delivered by the system (may be incomplete or absent)
+and the apps themselves as device specific APK files as installed at time of backup.
+
+*Storage ID*: A storage ID is the SHA-256 hash of the content stored in the repository.
```
t-m-w, line 31: [Clarity] The storage ID is basically the file name of any file in the repository, right? One that is encrypted so that it is not meaningful to those with access to the Seedvault data files, but lacking the keys? (A mapping of chunk IDs to storage IDs is kept in the snapshot files, which are encrypted.) If that's the case, consider using the term "file name" somewhere in this definition, just to make it all the more obvious. It's "required in order to load the file" because *it is the name of the file*, right?

I know it's described in the Repository Layout section below as well, but it's somewhat more roundabout than necessary. Somehow, it wasn't clear to me that this was the sole purpose of storage IDs (at least, I think that's the case?) until I read lines 326-327 in the Resume Interrupted Backup section.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+For restore, we iterate through all repositories and only offer snapshots for restore
+that can be decrypted with the key provided by the user.
+Hence, a restore may run on a second device while the first device is doing a backup.
+
+Note: A repository used for file backup is stored in a folder of the format `[ANDROID_ID].sv`
+and thus completely separate.
+
+## Data Format
+
+All files stored in the repository start with a version byte
```
t-m-w, line 102: [Clarity] Does "files" refer to blobs, or to truly all files in the repository (e.g. snapshots too)? Are there other files besides blobs and snapshots?

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+followed by an encrypted and authenticated payload (see also [Cryptography](#cryptography)).
```
t-m-w, line 103: [Clarity] Later (line 271), it is written that only the backup version byte is authenticated, along with an explanation of why that is the case. Until reaching that line of this document, it sounds like all of the data is authenticated. I don't know how much it matters, but it might be helpful to clarify that earlier (like around here) somehow.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+The version (currently `0x02`) is used to be able to modify aspects of the design in the future
+and to provide backwards compatibility.
+
+Blobs include the raw bytes of the compressed chunks
```
t-m-w: [Clarity] Nit: Blob payloads and snapshot payloads?

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+    "c11c3cf1c326375b177e768de908e6e423c3d8866715c8ec7159156b6ea82497": {
+      "id": "5421895dbead0ba83f0993f17b5f72f6c6618c0e9ea849c6c53ac240d1802d0b",
+      "length": 523496,
+      "uncompressedLength": 645314
+    }
+  }
+}
+```
+
+The `chunkIds` and `iconChunkIds` fields contain a list with plain text SHA-256 hashes
```
t-m-w: [Clarity] It's probably self-explanatory, but consider adding a short description of what iconChunkIds refers to.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+This section is based on and thus very similar to encryption of
+[files backup](../storage/doc/design.md).
+
+### Main Key
+
+Seedvault already uses [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki)
+to give users a mnemonic recovery code and for generating deterministic keys.
+The derived key has 512 bits
+and Seedvault used to use the first 256 bits as an AES key to encrypt app data.
```
t-m-w: 1. [Clarity] Nit: `s/used to use/previously used/`
2. [Information] Later (line 222), it is written that this first part of the key is no longer used. Maybe also mention that, or expand on it a bit, here. Is it retained for compatibility? Are there any important implications to no longer using this part of the key? etc. If the next line is intended to explain why it's not used, that is not immediately clear.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+Unfortunately, this key's usage is limited by Android's keystore to encryption and decryption.
+Therefore, the second 256 bits are be imported into Android's keystore for use with `HMAC-SHA256`,
```
t-m-w: 1. [Grammar] `s/are be imported/are imported/`
2. [Information] Were the second 256 bits unused in the past?

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+so that this key can act as a main key we can deterministically derive additional keys from
+by using HKDF ([RFC5869](https://tools.ietf.org/html/rfc5869)).
+These second 256 bits must not be used for any other purpose in the future.
```
t-m-w: [Emphasis] Nit: This may or may not be important enough to warrant additional emphasis somehow, e.g. bold, capital letters, something.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+This should be fine as the input key material is already a cryptographically strong key
+(see section 3.3 of RFC 5869 above).
+
+### Key derivation overview
+
+The original entropy comes from a BIP39 seed (12 words = 128 bit size)
+obtained from Java's `SecureRandom`.
+A PBKDF SHA512 based derivation defined in BIP39 turns this into a 512 bit seed key.
+
+The derived seed key (512 bit size) gets split into two parts:
```
t-m-w, line 250: [Clarity] Nit: Some of this section seems to duplicate the initial Main Key section. Consider adding a phrase like "As referenced earlier" to ensure that the reader easily understands that this section is discussing the same thing, rather than something new or subtly different.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+1. legacy app data encryption key (unused) - 256 bit - first half of seed key
+2. main key - 256 bit - second half of seed key used to derive application specific keys:
+   1. App Backup: HKDF with info "app backup repoId key"
+   2. App Backup: HKDF with info "app backup gear table key"
+   3. App Backup: HKDF with info "app backup stream key"
+   4. File Backup: HKDF with info "stream key"
+   5. File Backup: HKDF with info "Chunk ID calculation"
+
+### Stream Encryption
```
t-m-w, line 259: [Clarity]
I suspect that a "stream" refers to an input or output stream that would typically always be backed by a file...? Is there an important distinction between "streams" and "files"? Should this document *always* be talking about streams instead of files, or vice versa?

The Data Format section refers to "files". This is our first introduction to "streams" (it's not in the definitions).

If the distinction is important or otherwise intentional, then as with the above comment, there appears to be some duplicate information on the lines below - in this case from the Data Format section - so consider acknowledging that it was referenced earlier. Or perhaps the sections should be merged in some way...?

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+When a stream is written to backup storage,
+it starts with a header consisting of a single byte indicating the backup format version
+(currently `0x02`) followed by the encrypted payload.
+
+All data written to backup storage will be encrypted with a fresh key
+to prevent issues with nonce/IV re-use of a single key.
+
+We derive a stream key from the main key
+by using HKDF's expand step with the UTF-8 byte representation of "app backup stream key"
```
t-m-w: [Clarity] Nit: `s/representation of/representation of the string/` to mitigate momentary confusion of folks like me, assuming that's accurate.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+as info input.
+This stream key is then used to derive a new key for each stream.
+
+Instead of encrypting, authenticating and segmenting a cleartext stream ourselves,
+we have chosen to employ the [tink library](https://github.com/tink-crypto/tink-java) for that task.
+Since it does not allow us to work with imported or derived keys,
```
t-m-w: [Information] Is there a security relevance to this detail, or any other reason to keep this in mind? Is it suboptimal in some important way to solely use its AesGcmHkdfStreaming? I'm not knowledgeable in this area, so I'm not sure why this is important.

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+"Online Authenticated-Encryption and its Nonce-Reuse Misuse-Resistance"
+([PDF](https://eprint.iacr.org/2015/189.pdf)).
+
+It adds its own 40 byte header consisting of header length (1 byte), salt and nonce prefix.
+Then it adds one or more segments, each up to 1 MB in size.
+All segments are encrypted with a fresh key that is derived by using HKDF
+on our stream key with another internal random salt (32 bytes) and associated data as info
+([documentation](https://github.com/google/tink/blob/v1.5.0/docs/WIRE-FORMAT.md#streaming-encryption)).
+
+All types of files written to backup storage have the following format:
```
t-m-w: [Clarity] If I'm understanding correctly, this is a more-detailed version of a diagram that was shown earlier, so maybe this could read "Below is a more-detailed diagram of the format of all files written to backup storage", or something nicer-sounding to that effect. (Same deal as a few of my other comments.)

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+  * if APK was backed up already, re-use list of plaintext hashes for current snapshot
+  * if APK was not yet backed up, put it (and its splits) through the chunker
+  * chunks already in the repository are not uploaded again, only their hash recorded
+  * new chunks get compressed, encrypted and hashed to determine their storage ID, then uploaded 
+* for each app (the system allows us to back up):
+  * app data we receive as a tar stream from the system gets chunked
+    * chunks already in the repository are not uploaded again, only their hash recorded
+    * new chunks get compressed, encrypted and hashed to determine their storage ID, then uploaded
+  * remember ordered list of chunk IDs for the app (and its APKs)
+* add newly packed chunks to local index
```
t-m-w: [Outdated?] Later (line 390 and beyond), it is written that pack files and indexes are not used. So what does this line and the lines below mean, or are they outdated?

doc/README.md
```diff
@@ -0,0 +1,421 @@
...
+  and re-use the existing blob from the mapped storage ID
+* all mappings will be added to snapshot that gets uploaded at the end of backup
+
+### Restore
+
+* download all snapshots and let the user choose one for restore
+* for all apps that were selected for restore:
+  * reinstall app from APKs assembled from fetched chunks (see below for details)
+  * look up chunk IDs for app from snapshot
+  * download blobs for the chunk IDs as specified in snapshot
```
t-m-w, line 379: [Suggestion/Missing Info] Since the version byte is the only authenticated piece, shouldn't it be verified that the downloaded blobs match their hashes? This was likely already planned but not explicitly mentioned...?

t-m-w (changes requested):
Most of my comments are "nit" or clarification territory. Overall this looks great! I'm very excited about it.

doc/README.md
grote: Right. They are most likely files and in all current backends, they are files. However, they could be something else in the backend.

doc/README.md
grote: Thanks, clarified.

doc/README.md
grote: Thanks, I used your suggestion.

doc/README.md
grote: > The storage ID is basically the file name of any file in the repository, right?

No, but it is the beginning of the file name.

>  If that's the case, consider using the term "file name" somewhere in this definition, just to make it all the more obvious.

ok, done.

doc/README.md
grote: > Does "files" refer to blobs, or to truly all files in the repository (e.g. snapshots too)? Are there other files besides blobs and snapshots?

Not blobs, but *all* files, i.e. blobs *and* snapshots.

doc/README.md
grote: The encrypted data is authenticated. What you refer to is the authenticated associated data (AAD). The version byte is used as AAD, because it would otherwise not be authenticated.

doc/README.md
grote: sure, why not.

doc/README.md
grote: > `s/used to use/previously used/`

Thanks, changed.

>  Is it retained for compatibility? Are there any important implications to no longer using this part of the key? 

It is just there in the keystore of existing implementations and due to its use limitations, we can't really do much with it. Not sure what else to write.

doc/README.md
grote: > [Grammar] s/are be imported/are imported/

Thanks, fixed!

> [Information] Were the second 256 bits unused in the past?

Yes, they've been used already as the same main key for files backup. See `Key derivation overview`.

doc/README.md
grote: > I suspect that a "stream" refers to an input or output stream that would typically always be backed by a file...? Is there an important distinction between "streams" and "files"? Should this document always be talking about streams instead of files, or vice versa?

No, this is Java thinking. Also streams in Java don't need to be about files at all. *stream* here is crypto lingo, see also https://en.wikipedia.org/wiki/Stream_cipher

> Should this document always be talking about streams instead of files, or vice versa?

No. But not sure how to add more clarity.

doc/README.md
grote: yeah it zooms into the crypto part more, but since it is far away from that other diagram, not sure how to phrase this. Also, a reader my not always go from top to bottom.

doc/README.md
grote: good catch. a left over from a time the design had still packs and indexes

doc/README.md
grote: > Since the version byte is the only authenticated piece,

Do you know how you got this impression? Maybe we can prevent other from getting that as well.

> shouldn't it be verified that the downloaded blobs match their hashes? This was likely already planned but not explicitly mentioned...?

right, it was left rather implicit. it always needs to happen when reading files, so not sure the high-level procedures is a good place to mention it. I'll find another place.

doc/README.md
t-m-w: Okay, based on that explanation, I think this is fine as-is.

doc/README.md
t-m-w: What about adding a line like this after the previous, zoomed-out diagram:

"The structure of the encrypted tink payload is explored further in Stream Encryption." (where Stream Encryption is a link)?

doc/README.md
t-m-w: I think I searched the doc for instances of "auth" and the AAD section was the only place it was discussed in detail. As someone not familiar very familiar with the crypto details, to me, that segment sounded like it was justifying why certain things didn't need to be authenticated *at all*.

doc/README.md
t-m-w: I know I just responded to the other comment, so sorry for bouncing around, but I think I misinterpreted "otherwise not be authenticated" as meaning, "we wouldn't bother with authentication at all if not for wanting to authenticate the version byte" (implying, because it's not needed otherwise). Might be an uncommon misinterpretation. Rephrasing to something like, "to ensure it is also authenticated" would help.

doc/README.md
grote: What we are using is not the recommended high-level mentioned in their documentation, but hooks into a lower level. It felt important to mention this, so a reviewer could check what we do here and ensure it is actually safe.

doc/README.md
grote: good call, done.

Issue #671, opened 2024-05-30
Allow restore after device set-up


grote:
* [x] potentially pre-uncheck system apps and settings
* [x] run in a foreground service in case user navigates away #203
* [x] in first fresh app start, ask if user wants to restore old backup or start fresh #571
* [x] installing of apps should only happen if not already installed (or on older version)
* [x] handle auto-restore setting (maybe temp disable?)
* [x] enable restore menu item out of dev builds
* [x] remove dialer code (from code and FAQ)
* [x] don't create duplicate files for storage/files restore

DDvO:
Hopefully this finally solves #85, which in my view got closed prematurely?

grote:
> Hopefully this finally solves #85, which in my view got closed prematurely?

It did get closed, because there is an (unsupported) way to access restore outside of SUW. This ticket is about properly supporting it and as you see there's quite a bit of work involved.

PR #718, opened 2024-08-15
Uncheck system apps by default if restore outside of SUW

grote:
Part of #671


t-m-w (approved):
Worked for me! Checked in SUW as expected, unchecked outside of it.

grote:
thanks for testing!

PR #719, opened 2024-08-15
Don't re-install apps that are already installed and disable auto-restore

grote:
This doesn't re-install apps that are already installed at the same or a newer version. Also, if auto-restore is enabled during install, it disables it and re-enables it when install is complete.

Part of #671

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/App.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/App.kt
@@ -17,7 +17,6 @@ import android.os.ServiceManager.getService
 import android.os.StrictMode
 import android.os.UserHandle
 import android.os.UserManager
-import android.provider.Settings
 import androidx.work.ExistingPeriodicWorkPolicy.UPDATE
 import androidx.work.WorkManager
 import com.google.android.material.color.DynamicColors
@@ -148,6 +147,7 @@ open class App : Application() {
     private val metadataManager: MetadataManager by inject()
     private val backupManager: IBackupManager by inject()
     private val pluginManager: StoragePluginManager by inject()
+    private val backupStateManager: BackupStateManager by inject()
 
     /**
      * The responsibility for the current token was moved to the [SettingsManager]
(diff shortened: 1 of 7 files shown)


t-m-w (commented):
This works! I have a couple of thoughts, but I'm not sure if they even ought to be addressed:
1. Apps that were already installed do not show app icons, just the generic icon. This looks a little odd alongside apps that were not already installed, as if maybe something went wrong.
2. Apps that were already installed appear so abruptly - especially if all the apps were already installed - that, as above, it looks like maybe something went wrong.

The apps do have green checkmarks, which indicate everything should be fine, but I could see how a user might get a mixed message from the observed behavior.

*Ideally*, it could be communicated somehow that the app was already installed, but I don't think it's necessarily a blocker.

t-m-w (dismissed):

grote:
I think at least the first point should get addressed. However, I can't reproduce it. See this video for how things look for me:

https://github.com/user-attachments/assets/262de625-7544-4fd7-a853-cd150f53b9e7

There's a bit of abruptness when all the apps suddenly show up, but doesn't look bad to me.

grote:
I added two commits related to disabling auto-restore during install to avoid restoring the same data twice for affected apps.

t-m-w:
> I think at least the first point should get addressed. However, I can't reproduce it.

We spoke about this, and there were a few things that came to mind:
1. My backup doesn't have APKs, so it couldn't pull icons or names from those.
2. My backup likely doesn't have icons either, although I'm not really sure how to confirm that. The initial restore selection screen doesn't show either icons or names - is that confirmation enough, maybe?

<details><summary>Screenshot of initial restore selection screen</summary>

Shows a lack of icons and names (just shows package name) for 4 apps.


</details>

So assuming point 2 above is correct, then this is an edge case not worth worrying much about.

t-m-w (approved):
From a user-facing perspective (no code review or adb, just what I can see happening), this all appears to work fine to me!

t-m-w:
>     2. My backup likely doesn't have icons either

I think this was all true. My backup was from a production build of CalyxOS, for which no stable version currently includes the change to include icons in the backups (should be in 5.10.x which is almost stable). So, it did not have icons or names stored.

I made another backup with the latest upstream code along with these changes, and the described/shown issue is not present. Icons and names are shown.

stevesoltys (approved):

PR #720, opened 2024-08-16
Fix app selection when restoring after SUW

grote:
Even when one item was unchecked, the top item said all checked which caused some selection logic errors.


theimpulson (approved):

t-m-w (approved):
Behavior is as described, and code looks good.

PR #722, opened 2024-08-20
Start a foreground service during restore

grote:
Part of https://github.com/seedvault-app/seedvault/issues/671


t-m-w (approved):
This seems to work fine. My test sets have been small, making for a short restore process, so I don't know for sure how it would handle a longer process. But for small sets at least, no problems encountered here.

PR #725, opened 2024-08-21
Don't back up to USB, if backup disabled

### LINKED ISSUE #724: APK backups occur even when "Backup my apps" is off with USB devices
APK backup is happening for USB devices even when "Backup my apps" is off. The "App backup" toggle is grayed out in this situation, so the user cannot adjust its state.

## Steps to reproduce
1. Start from a clean install, or at least from a state where the "Backup my apps" toggle is not turned on
2. Either set up a new Seedvault backup to a USB device, or restore an existing backup from a USB device
3. (Optional, but useful) In the system's Settings > Notifications, turn on Notification history
4. Eject the USB device
5. Reconnect the USB device

## Expected result
No modification of the backup storage is attempted when "Backup my apps" is off (unless Storage backups are on, but aside from that).

## Actual result
A notification will appear about APK backups. Look in Notification history in case you blinked and missed it.

`adb logcat` will include lines like these:
```
I AppBackupWorker: Starting APK backup... (stopped: false)
I AppBackupWorker: Asking to do app backup now...
```

grote:
Fixes #724

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/UsbIntentReceiver.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/UsbIntentReceiver.kt
@@ -5,6 +5,7 @@
 
 package com.stevesoltys.seedvault
 
+import android.app.backup.IBackupManager
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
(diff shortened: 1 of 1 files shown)


t-m-w (approved):
Looks good and appears to work.

PR #734, opened 2024-08-29
Unify Storage Backends

grote:
We had two storage plugin interfaces: One for app backup and one for files backup. Therefore we had two different WebDAV implementations and two different Storage Access Framework implementations. The storage interface was application specific to its use-case.

This MR unifies those implementations using a single storage backend interface which is rather generic and mostly independent of application specific use cases:

``` kotlin
public interface Backend {
    public suspend fun test(): Boolean
    public suspend fun getFreeSpace(): Long?
    public suspend fun save(handle: FileHandle): OutputStream
    public suspend fun load(handle: FileHandle): InputStream
    public suspend fun list(topLevelFolder: TopLevelFolder?, vararg fileTypes: KClass<out FileHandle>, callback: (FileInfo) -> Unit)
    public suspend fun remove(handle: FileHandle)
    public suspend fun rename(from: TopLevelFolder, to: TopLevelFolder)
    public val providerPackageName: String?
}
```

This should make writing new storage backends easier. It reduces lots of duplicated code and allows for future modernizations such as retrying operations after I/O errors. 

The new unified implementations also improve upon their old separate implementations. There are no more `init()` or `startRestoreSet()` methods where we created new folders. Instead, all folders are now created on-demand, only when they are needed and in the case of SAF also cached for faster access. Initializing a new backup location is faster, because we don't need to create the 255 folders for file backup.

The implementation now lives in a new `core` gradle module, so the `app` and the `storage` module can both make use of it. In the future, common cryptographic code should also get moved there to further reduce duplication.


t-m-w:
I've been running this on my daily driver and haven't encountered problems. I've been seeing notifications of successful backups daily. I tested restoring a single app after making a small change - a notes app - and it appears to work fine. Haven't done full-blown testing, but haven't encountered errors, either.

I use WebDAV as a backend.

t-m-w (approved):
Seems fine to me as per my last comment.

PR #736, opened 2024-09-05
Fix retries: don't upload metadata three times

grote:


t-m-w (approved):
Not tested yet, but code changes look good to me. (I accidentally manifested this same kind of problem in a script just the other day, forgetting to break out of a retry loop...)

PR #742, opened 2024-09-09
Android 15 updates

theimpulson:


gradle/libs.versions.toml
```diff
@@ -5,14 +5,14 @@
...
+androidGradlePlugin = "8.6.0"
 protobufPlugin = "0.9.4"
 ktlint = "11.5.0"
 
 # Android SDK versions
-compileSdk = "34"
-minSdk = "34"
-targetSdk = "34"
+compileSdk = "35"
+minSdk = "35"
```
grote: Could we leave this at `34` for now please?

grote (changes requested):
Awesome, thanks a lot!

Was `contactsbackup/libs/com.android.vcard.jar` still in the same place or do we need to update the comment?

gradle/libs.versions.toml
theimpulson: Done

theimpulson:
> Awesome, thanks a lot!
> 
> Was `contactsbackup/libs/com.android.vcard.jar` still in the same place or do we need to update the comment?

It was still at the same place but I had to run `m com.android.vcard` command manually to build it.

grote:
ARgh: `IllegalArgumentException: failed to configure com.stevesoltys.seedvault.restore.install.DeviceInfoTest.test supported languages: Package targetSdkVersion=35 > maxSdkVersion=34`

What targetSdk does the AOSP build set? is there a way we set target to 34 for now? Or can we change target only for tests?

theimpulson:
> ARgh: `IllegalArgumentException: failed to configure com.stevesoltys.seedvault.restore.install.DeviceInfoTest.test supported languages: Package targetSdkVersion=35 > maxSdkVersion=34`
> 
> What targetSdk does the AOSP build set? is there a way we set target to 34 for now? Or can we change target only for tests?

Done

grote (approved):
awesome, thanks!

PR #749, opened 2024-09-19
Navigation Improvements

theimpulson:


grote:
i see something strange happening here when navigating back from fragments. do you see the same?


https://github.com/user-attachments/assets/a315369c-0524-43da-8701-61dd43d0a6d2

Note the toolbar with title and overflow menu.

theimpulson:
> i see something strange happening here when navigating back from fragments. do you see the same?

Seems fine on Android 15 here

storage/demo/src/main/java/de/grobox/storagebackuptester/restore/DemoSnapshotFragment.kt
```diff
@@ -31,7 +31,7 @@ class DemoSnapshotFragment : SnapshotFragment() {
         bottomStub.layoutResource = R.layout.footer_snapshot
         val footer = bottomStub.inflate()
         footer.findViewById<Button>(R.id.button).setOnClickListener {
-            requireActivity().onBackPressed()
+            requireActivity().onBackPressedDispatcher.onBackPressed()
```
grote, line 34: If you change this for the demo app, don't you need to add `android:enableOnBackInvokedCallback="true"` to its manifest as well?

theimpulson: Done

grote:
Still seeing the same thing on Android 15 here.

PR #767, opened 2024-10-03
Fix title of restore notification channel

t-m-w:
The restore foreground service notification uses the restore channel, which was inadvertently using the title from the error channel, showing "Auto restore flash drive error" when tapping and holding on the notification, and showing that channel name twice in the notification settings.


grote:
Thanks!

grote (approved):

PR #772, opened 2024-10-07
Also show TikTok in Backup status

### LINKED ISSUE #769: TikTok doesn't show up in the app list, but is backed up anyway
I have [TikTok](https://play.google.com/store/apps/details?id=com.zhiliaoapp.musically) installed, for some reason when going into Seedvault > Backup status, it cannot be found. It is also not possible to remove it from being backed up.

TODO: try and get logs

OS: CalyxOS 5.11.2
Seedvault version: 14-4.1

grote:
The apps ships an instrumentation configuration which we've so far only soon for instrumentation test dev apps. Now, we do more checks to identify those.

Fixes #769


t-m-w (approved):
Untested but LGTM

PR #776, opened 2024-10-13
Switch to toolbar APIs for working with menu

theimpulson:
## Summary

Activity methods have been deprecated

### DIFF
--- a/storage/demo/src/main/res/layout/fragment_scan.xml
+++ b/storage/demo/src/main/res/layout/fragment_scan.xml
@@ -14,7 +14,6 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:minHeight="?attr/actionBarSize"
-        android:theme="?attr/actionBarTheme"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent"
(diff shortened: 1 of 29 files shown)


storage/demo/src/main/res/layout/fragment_scan.xml
```diff
@@ -9,13 +9,24 @@
...
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:menu="@menu/fragment_scan"
+        app:navigationIcon="?attr/homeAsUpIndicator" />
+
     <androidx.core.widget.NestedScrollView
         android:id="@+id/scrollView"
         android:layout_width="match_parent"
-        android:layout_height="match_parent"
+        android:layout_height="0dp"
```
grote, line 26: do you know why this works? There's only a top constraint and no bottom. So why doesn't this shrink it to actual height `0`?

grote (approved):
Thanks, looks good now.

storage/demo/src/main/res/layout/fragment_scan.xml
theimpulson: @grote not sure. Do you want me to put a bottom constraint here though?

grote: It *does* seem to work, so this shouldn't be needed, I just wonder how this can even work :shrug:

PR #786, opened 2024-10-30
Show 'Launch' button for force stopped apps in backup status screen

grote:


t-m-w:
This seems to work just fine, except for apps that by design cannot be launched:



An optional very-wishlist item would be a way to indicate to the user that the app is no longer stopped (if we even know) and should be backed up on the next run.

t-m-w (approved):
Launched some apps that hadn't been backed up, using this new button. Ran backup again. They're all good now.

grote:
> This seems to work just fine, except for apps that by design cannot be launched

I wonder why these apps show up there at all. The expectation is that the only system apps it shows there are the launchable ones.

grote:
> An optional very-wishlist item would be a way to indicate to the user that the app is no longer stopped (if we even know) and should be backed up on the next run.

Sounds worthy of a ticket :P

t-m-w:
> > An optional very-wishlist item would be a way to indicate to the user that the app is no longer stopped (if we even know) and should be backed up on the next run.
> 
> Sounds worthy of a ticket :P

All right, done here: https://github.com/seedvault-app/seedvault/issues/788

PR #785, opened 2024-10-30
Verify app backup integrity

grote:
Adds a new option to app backup section in settings which brings up a new screen for checking app backups. User can define how large the random sample to be checked shall be, from 5% to 100%. Then checking happens in the background reporting progress in a notification. A final notification gets shown when the process finished when brings up a detailed report about the issues found, if any. Usually, a success screen is shown.

Here the error case:

https://github.com/user-attachments/assets/6a490e3e-4513-40db-9c23-d748468b3aa5

Fixes #290, #798

PR #790, opened 2024-11-07
Switch from sharedViewModel to activityViewModel

theimpulson:
sharedViewModel was deprecated in favor of activityViewModel

Ref: https://insert-koin.io/docs/reference/koin-android/viewmodel/#activity-shared-viewmodel


grote:
Thanks, but I think this is already part of https://github.com/seedvault-app/seedvault/pull/785/commits/f67c1d5544d1452e4c2a83ac7945190fb680c10f

theimpulson:
Ah I missed that. Nice.

PR #803, opened 2024-11-23
README: Add link on where to translate SeedVault

lucasmz-dev:
Not pointing to the specific Seedvault component as there is also two other components for Seedvault


grote (approved):
Thanks!

PR #810, opened 2024-12-09
Create SECURITY.md

grote:


SECURITY.md
```diff
@@ -0,0 +1,18 @@
+# Reporting Security Issues
+
+The Seedvault team and community take security bugs seriously.
+We appreciate your efforts to responsibly disclose your findings, and will make every effort to acknowledge your contributions.
+
+To report a security issue,
+please use the GitHub Security Advisory ["Report a Vulnerability"](https://github.com/seedvault-app/seedvault/security/advisories/new) tab.
```
chirayudesai: security@seedvault.app is live now.

I think we could put that up as an option if somebody doesn't want to use GitHub. It'd be nice to have everything in one place otherwise.

chirayudesai (commented):
Looks good!

SECURITY.md
grote: added

chirayudesai (approved):

PR #812, opened 2024-12-11
Add untranslatable flag to  "backup_app_check_error_title" string

lucasmz-dev:
Shows up as a translatable string on Weblate
(could be dealt there only but this is nicer)


grote (approved):
Thanks, good catch!

grote:
Note that weblate still points to android14, we need to switch to 15 still. so the change will not show up there immediately.

Issue #805, opened 2024-11-28
Backup location disabled after backup failed on re-formated flash drive


An-anonymous-coder:
After GrapheneOS updated Seedvault in their [latest release](https://grapheneos.org/releases#2024112700), I'm running into an issue. I'm going to give a rough outline of what may have caused it:

1. Format a new USB drive for Android storage. The name of it is "android", I don't know if there is a way to change this.
2. Using the old version of Seedvault, create a backup location inside the drive
3. Backup like normal
4. Eject the drive
5. Update Seedvault to the new version
6. Insert the USB drive
7. Format the USB drive. This will wipe the Seedvault data and the device will be named "android" still.
8. Open Seedvault

You will receive a notification "**Backup failed:** An error occurred while running the backup"

The option to select the backup location is grayed out, even when the USB drive is not plugged in.

Backup scheduling is set to 1 week. Backup status says "Last backup: Never", even though it has been backed up recently. 

I essentially have no way to backup to this USB drive now, or anywhere else since I can't change the backup location. 

```
version 15-5.1
versionCode 35050010
Android 15: GrapheneOS 2024112700
```

An-anonymous-coder:
I like to frequently format the USB drive for backup sanity and cleanliness, especially after major updates like this.

An-anonymous-coder:
I tried formatting the drive off-device and naming it something other than "android". Same error message, still no way to change backup location. 

Should I clear Seedvault's cache and storage?

grote:
Can you reproduce the issue using steps 6 to 8 only? I just want to be sure that the upgrade to the new version is not involved here.

Otherwise, frequently reformatted the backup medium isn't something we really tested so far.

> Should I clear Seedvault's cache and storage?

Yes, if there's no data to lose, you can do that, also to get back to a working state to reproduce 6-8 above.

If you can reproduce, and after formatting, backup fails and you can't change storage location, then please export logs from expert settings and post them here. Thanks.

An-anonymous-coder:
I cleared cache and storage and I was able to backup as normal.

Repeating steps 6-8 reproduced the bug. I didn't receive an error notification. It says the last backup was (correctly) 5 minutes ago. Here is a revised bug report:

1. Format a new USB drive for Android storage. The name of it is "android", I don't know if there is a way to change this.
2. Create a backup location inside the drive
3. Backup like normal.
After the backup is finished, you may eject and reinstert the flash drive, if desired. This isn't necessary. 
[...]

PR #819, opened 2024-12-16
Integrity verification of files backup

### LINKED ISSUE #567: Integrity verification of files backup
Currently, there is no way to verify the integrity of files backup without doing an actual restore (on another device) and manually comparing the restored data. In this task, we want to offer an automatic verification mechanism to give users peace of mind.

* [ ] Implement way to download and verify backup integrity without actually restoring anything
* [ ] Show how much data of each folder and how many files got backed app

grote:
This implements a way to download and verify backup integrity without actually restoring anything. It also show how much data of each folder and how many files got backed up.

The PR adds a new option to file backup section in settings which brings up a new screen for checking file backups. User can define how large the random sample to be checked shall be, from 5% to 100%. Then checking happens in the background reporting progress in a notification. A final notification gets shown when the process finished. It brings up a detailed report about the issues found, if any. Usually, a success screen is shown.

If backups could be decrypted, the user can tap individual backups and see their contents, which files are in there, what is their size and timestamp. If a backup corruption happened, it is shown which files are affected by it.

Here the error case:

https://github.com/user-attachments/assets/181e2cd4-7195-48ce-98f2-ec0405e11b7b

Here only one corrupted backup with the rest being fine:

https://github.com/user-attachments/assets/479d7525-4931-4847-b5d7-802797819edd

Success case:

https://github.com/user-attachments/assets/06a3748c-2c97-4f7a-af17-08f571691081

All backups corrupted:

Screenshot_20241218_091221

Closes #567, #805


t-m-w (approved):
Test (repeated separately for `printf '\0' >> "$file"`, `truncate -s -1 "$file"`, `cat "$file" | tr '[abcdef]' '[fedcba]' > "${file}2"; rm "$file"; mv "${file}2" "$file"`, and `rm "$file"`:
1. Change a file in the `.sv` subdir in one of the aforementioned ways.
2. Run an integrity check. It should fail and point to one or more files as being corrupted.
3. Run a backup.
4. Run an integrity check. If a new snapshot is created (when? i.e. if it's been long enough?), the integrity check should fail, but the latest snapshot in the results list will be fine. If a new snapshot is not created, it will succeed and existing files will be repaired/re-created.

When I first tried this, this morning, it created a new snapshot. But I removed the files backup directory and ever since it kept using the same snapshot.

I *think* this is all working as intended?

Additionally, the "Backup now", "Check integrity", and "Restore backup" buttons / menu items become grayed out during an integrity check. My integrity check happens very fast with only 30MB of files, so I couldn't confirm if it stayed this way long-term (i.e. if it doesn't undo the grayed-ness early), but at least with my testing, it all looked fine.

PR #821, opened 2024-12-16
Move context SAF cross-profile hack

grote:
so it can be used before we save SafProperties, e.g. during restore when we need to check if backups at the given location exist.


grote:
@Uldiniad this is untested. Please also test the normal in profile usb and internal storage use-case, so those don't break by this change.

Uldiniad:
> @Uldiniad this is untested. Please also test the normal in profile usb and internal storage use-case, so those don't break by this change.

Tested internal storage and USB restore on main user and work profile. Everything worked as expected

Uldiniad (approved):

Uldiniad (approved):

PR #668, opened 2024-05-30
Translations update from Hosted Weblate

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [CalyxOS/Seedvault](https://hosted.weblate.org/projects/calyxos/seedvault/).

It also includes following components:

* [CalyxOS/Seedvault contacts backup](https://hosted.weblate.org/projects/calyxos/seedvault_contactsbackup/)

* [CalyxOS/Seedvault storage backup](https://hosted.weblate.org/projects/calyxos/seedvault_storage/)

Current translation status:

Weblate translation status

PR #836, opened 2025-01-04
Cherry-pick translations from android14 weblate

lucasmz-dev:
#668 but for `android15`

Weblate is locked so this should be safe (https://hosted.weblate.org/projects/calyxos/seedvault/#information)

CC: @chirayudesai @mikeNG


chirayudesai:
Another option: Directly edit target branch of https://github.com/seedvault-app/seedvault/pull/668 - but I don't know if that'll mess anything up upstream.

Not merging it to 14 directly since the translations will end up there anyways via the backports.

lucasmz-dev:
@chirayudesai doesn't work well:
https://github.com/lucasmz-dev/CONTRIB-seedvault/pull/1

merges extra code reverts and changes and can't be done without dealing with conflicts

PR #850, opened 2025-01-18
strings: use APK wording instead of app

lucasmz-dev:
This setting is now in expert settings, maybe we can be more specific?

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -28,12 +28,12 @@
     <string name="settings_auto_restore_summary">When reinstalling an app, restore backed up settings and data.</string>
     <string name="settings_auto_restore_summary_usb">Note: Your %1$s needs to be plugged in for this to work.</string>
     <string name="settings_category_app_data_backup">App data backup</string>
-    <string name="settings_backup_apk_title">App backup</string>
-    <string name="settings_backup_apk_summary">Back up the apps themselves. Otherwise, only app data would get backed up.</string>
-    <string name="settings_backup_apk_dialog_title">Really disable app backup?</string>
-    <string name="settings_backup_apk_dialog_message">Disabled app backup will still back up app data. However, it will not get restored automatically.\n\nYou will need to install all your apps manually while having \"Automatic Restore\" switched on.</string>
+    <string name="settings_backup_apk_title">APK backups</string>
+    <string name="settings_backup_apk_summary">Backs up the apps themselves. Otherwise, only app data will get backed up.</string>
+    <string name="settings_backup_apk_dialog_title">Really disable APK backups?</string>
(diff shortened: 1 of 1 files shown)


chirayudesai:
Yes, I like this. I think "APK" is a standard enough Android term, and also makes it less confusing.

grote (dismissed):

app/src/main/res/values/strings.xml
```diff
@@ -28,12 +28,12 @@
     <string name="settings_auto_restore_summary">When reinstalling an app, restore backed up settings and data.</string>
     <string name="settings_auto_restore_summary_usb">Note: Your %1$s needs to be plugged in for this to work.</string>
     <string name="settings_category_app_data_backup">App data backup</string>
-    <string name="settings_backup_apk_title">App backup</string>
+    <string name="settings_backup_apk_title">APK backup</string>
     <string name="settings_backup_apk_summary">Back up the apps themselves. Otherwise, only app data would get backed up.</string>
```
chirayudesai: maybe change this too?

app/src/main/res/values/strings.xml
```diff
@@ -28,12 +28,12 @@
...
-    <string name="settings_backup_apk_dialog_title">Really disable app backup?</string>
-    <string name="settings_backup_apk_dialog_message">Disabled app backup will still back up app data. However, it will not get restored automatically.\n\nYou will need to install all your apps manually while having \"Automatic Restore\" switched on.</string>
+    <string name="settings_backup_apk_dialog_title">Really disable APK backup?</string>
+    <string name="settings_backup_apk_dialog_message">Disabled APK backup will still back up app data. However, it will not get restored automatically.\n\nYou will need to install all your apps manually while having \"Automatic Restore\" switched on.</string>
```
chirayudesai: Unrelated nit, "Disabled" could be "Disabling" instead.

chirayudesai (changes requested):

app/src/main/res/values/strings.xml
lucasmz-dev: I feel like that description is useful, it implies "APK" is the app itself and that may be useful

app/src/main/res/values/strings.xml
lucasmz-dev: I think overall this could be improved, instead of a simple word replace, e.g. "When disabling APK backups, app data will still be backed up..."

I can do this, but I'd rather do this in another PR

lucasmz-dev: I did this

cde541 (approved):

chirayudesai (approved):

grote:
Thanks lucas!

PR #854, opened 2025-01-20
Create codacy.yml to see if their security scanner is helpful

grote:


grote:
This mostly found things with shellcheck which isn't too bad as auditors run this as well: https://github.com/seedvault-app/seedvault/security/code-scanning?query=pr%3A854+is%3Aopen

PR #855, opened 2025-01-20
Create mobsf.yml

grote:


grote:
I am not impressed by https://github.com/seedvault-app/seedvault/security/code-scanning?query=pr%3A855+is%3Aopen but since pentesters use the same thing, maybe worth to include, so we can proactively react to those issues.

Issue #822, opened 2024-12-16
Set a higher quota for app backups


John-Rodriguez-Git:
Im a heavy cromite user and a tab junkie/messie and have over 3GB User Data. I backup to my 512GB nvme and would want to set a higher quota.
Why was 
https://github.com/seedvault-app/seedvault/pull/750/commits/03d2946c93813602c0efd0d7d80b9d0a5852f367
done?

Linking to https://github.com/seedvault-app/seedvault/issues/183 and https://github.com/seedvault-app/seedvault/issues/231 as the described solution to increase the quota is not available anymore since that commit.

grote:
I don't think we should bring back the option to manually set the quota. Instead, we could bump the quota to a higher value, 5 GB maybe.

Waffle42:
Please bring it back.

oxalica:
I updated my phone system from LineageOS 21.0 to 22.1 (major update) yesterday and did a storage reset for clean installation. It turns out some of my apps failed to restore due to "quote exceeded".
The most painful one is [fcitx5-android](https://github.com/fcitx5-android/fcitx5-android), the Chinese input method which contains my own manually-trained dictionary. It is 53MB in the backup but failed to restore due to quota issue. I guess I turned on "unlimited quota" on backup, but now that option is gone from expert mode. How can I restore my own existing data now?

grote:
This ticket is unrelated to restore. If restore failed due to quota exceeded, then you did the backup with an old version and limited quota. The current quota of the recent version can handle 53MB just fine.

Waffle42:
> I don't think we should bring back the option to manually set the quota. Instead, we could bump the quota to a higher value, 5 GB maybe.

Is there a reason for why you wouldn't want to bring it back? Shouldn't the power be held by the user?

oxalica:
> This ticket is unrelated to restore. If restore failed due to quota exceeded, then you did the backup with an old version and limited quota. The current quota of the recent version can handle 53MB just fine.

Okay maybe I misunderstood the issue. I'm on seedvault 15-5.2 which should contain the 1GB default change mentioned in the issue description. But in "Restore backup" I click the 2 days ago backup, and Fcitx5 shows "Last backup: December 10, 2024 (52MB)". When I select it and click "Restore backup", then it shows "Quota exceeded". Do you mean the first information is incorrect and the app is not actually backed up to my remote storage? This is also misleading to me...

grote:
[...]

PR #857, opened 2025-01-22
Several assorted improvements

### LINKED ISSUE #825: Webdav backend incompatible with golang stdlib webdav implementation
I think I found conflict between how the seedvault implementation expects directories to work and how golang handles it:

In this function, you test that directories exist using the HEAD function:
https://github.com/seedvault-app/seedvault/blob/156ffbd8bc7d8e9f35472454fb00cb6c0a167aee/core/src/main/java/org/calyxos/seedvault/core/backends/webdav/WebDavHelper.kt#L84-L103

golang doesn't support HEAD requests on directory listings:
https://github.com/golang/net/blob/dfc720dfe0cfc125116068c20efcdcb5e4eab464/webdav/webdav.go#L212

I checked the RFCs [RFC4918](http://www.webdav.org/specs/rfc4918.html#rfc.section.9.4) and [RFC2068](https://datatracker.ietf.org/doc/html/rfc2068#section-9.4) but couldn't find any defintive answer how this should be handled. But there isn't really an answer to this, beyond "Treat GET and HEAD identical". The overall internet also seems undecided how to handle directory requests in webdav. ([Here one of the samples I found](https://stackoverflow.com/a/24338433)).

I think the correct way to check for directories would be the usage of `PROPFIND` with a depth of 0. That would certainly be in line with the [RFC4918 section 9.1](http://www.webdav.org/specs/rfc4918.html#rfc.section.9.1).

This is an upstream issue for downstream issue https://github.com/GrapheneOS/os-issue-tracker/issues/4558

grote:
* bump app quota to 3GiB
* use propfind instead of head
* fix error message when there are no backups to restore
* trust user CA certs

Closes #825 #822

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
   SPDX-FileCopyrightText: 2017-2019 Steve Soltys
   SPDX-FileCopyrightText: 2020 The Calyx Institute
   SPDX-License-Identifier: Apache-2.0
@@ -73,7 +72,8 @@
         tools:ignore="ProtectedPermissions" />
 
     <!-- Used to get the name of the current profile -->
-    <uses-permission android:name="android.permission.QUERY_USERS"
+    <uses-permission
+        android:name="android.permission.QUERY_USERS"
         tools:ignore="ProtectedPermissions" />
 
     <!-- Used for periodic storage backups -->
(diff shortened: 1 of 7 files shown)


t-m-w:
Smoke test is fine. This code has been active on my daily driver for 6 days with no visible issues. Nightly backups have been succeeding.

Would you like me to test the functionality, e.g. user CA cert?

grote:
> Would you like me to test the functionality, e.g. user CA cert?

would be nice, but not essential. I had tested this myself. Maybe it makes a difference with an AOSP build?

t-m-w:
> would be nice, but not essential. I had tested this myself. Maybe it makes a difference with an AOSP build?

I'm going to accept the "not essential" aspect of this and go ahead and approve!

t-m-w (approved):
Passes smoke test for the last few weeks, seems fine.

PR #873, opened 2025-02-16
Fix a few issues with strings

lucasmz-dev:
* Be consistent in using "file backup" instead of "storage backup" (following 41f47620e4044be6b3e5eed58f3c237e4c9e7cf6)
* Fix a few strings to be marked (or not anymore) as not translatable


grote (approved):
Thanks!

PR #874, opened 2025-02-17
Use different allow-lists for backup and restore

grote:


app/src/main/res/values/config.xml
```diff
@@ -1,14 +1,25 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
```
chirayudesai, line 1: Heh that looks ok, but it builds fine so I guess that's ok...

chirayudesai (approved):
LGTM!

PR #875, opened 2025-02-17
Update design.md to reflect #867

### LINKED ISSUE #867: File backup data gets corrupted on an interrupted nextcloud backup
Every time the backup process for storage starts, it says "Scanning for files..." meanwhile deleting everything available on the nextcloud server which contains the previous backup. The feature supporting reusing of app backups appears to be working fine, so I assume that it is probably intended for it to work the same for storage backup at some point.

Seedvault version: 15-5.2 on GrapheneOS

EnforcerE:
close #867

### DIFF
--- a/storage/doc/design.md
+++ b/storage/doc/design.md
@@ -420,6 +420,7 @@ but are considered out-of-scope of the current design for time and budget reason
 * external secret-less corruption checks that would use checksums over encrypted data
 * concealing file sizes (though zip chunks helps a bit here)
 * implementing different storage plugins
+* being able to recover data from a partial backup
 
 # Known issues
 


storage/doc/design.md
```diff
@@ -420,6 +420,7 @@ but are considered out-of-scope of the current design for time and budget reason
 * external secret-less corruption checks that would use checksums over encrypted data
 * concealing file sizes (though zip chunks helps a bit here)
 * implementing different storage plugins
+* being able to recover data from a partial backup
```
grote, line 423: this is a bit tricky, because there are no partial backups, so you could say:

* making and persisting incomplete backups

PR #884, opened 2025-03-07
Enable nonTransitiveRClass and fix breakage it causes

grote:


grote:
i did a quick test by throwing a fake exception to cause the error and the color looked OK in light and dark theme.

cde541 (changes requested):
Fails to build in AOSP,

```
packages/apps/Seedvault/storage/lib/src/main/java/org/calyxos/backup/storage/ui/restore/SnapshotFragment.kt:61:75: error: unresolved reference: attr
                    val color = MaterialColors.getColor(emptyStateView, R.attr.colorError)
```

grote:
thanks for testing, but also strange. it builds fine on QPR1, why would it not find `R.attr` which is a normal thing since forever.

chirayudesai:
https://github.com/seedvault-app/seedvault/blob/aa4468bf9582f5e6a53fd8df510171897a00b5f0/app/src/main/java/com/stevesoltys/seedvault/restore/RestoreSetAdapter.kt#L87

chirayudesai:
TODO: Enable https://github.com/seedvault-app/seedvault/blob/android15/gradle.properties#L12

grote:
should  build in AOSP now, just tested it

chirayudesai:
> should build in AOSP now, just tested it

It does, thanks!

chirayudesai (approved):

PR #885, opened 2025-03-07
Simplify debug build check logic

herobuxx:
Since there only one non-debugging build variant, i think it's better to keep it simple

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/App.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/App.kt
@@ -193,7 +193,7 @@ const val ERROR_BACKUP_CANCELLED: Int = BackupManager.ERROR_BACKUP_CANCELLED
 const val ERROR_BACKUP_NOT_ALLOWED: Int = BackupManager.ERROR_BACKUP_NOT_ALLOWED
 
 // TODO this doesn't work for LineageOS as they do public debug builds
-fun isDebugBuild() = Build.TYPE == "userdebug" || Build.TYPE == "eng"
+fun isDebugBuild() = Build.TYPE != "user"
 
 fun <T> permitDiskReads(func: () -> T): T {
     return if (isDebugBuild()) {


app/src/main/java/com/stevesoltys/seedvault/App.kt
```diff
@@ -193,7 +193,7 @@ const val ERROR_BACKUP_CANCELLED: Int = BackupManager.ERROR_BACKUP_CANCELLED
 const val ERROR_BACKUP_NOT_ALLOWED: Int = BackupManager.ERROR_BACKUP_NOT_ALLOWED
 
 // TODO this doesn't work for LineageOS as they do public debug builds
-fun isDebugBuild() = Build.TYPE == "userdebug" || Build.TYPE == "eng"
+fun isDebugBuild() = Build.TYPE != "user"
```
grote, line 196: I don't know enough about possible build types, but it seems to me making debug builds explicit is safer than declaring everything a debug build which isn't `user`. Also considering that certain safety features are disabled in debug builds.

herobuxx: There are only three build variants: user, userdebug, and eng.

Since userdebug and eng are both debugging builds, classifying any variant other than user as a debugging build would simplify code readability.

grote: what if there is ever a new build variant introduced?

mikeNG: Only those 3 variants have existed in the ~17 years of android being public.

I am not sure if this change is worth making, it doesn't actually fix something and the TODO above is still valid, LineageOS and other downstream forks will still have to modify this check to just  `Build.TYPE == "eng"`

grote: Thanks, I agree not worth making this change.

Issue #636, opened 2024-03-21
New Profile Backup restore: No Back Button on wrong entered Passphrase-Words


TimoGoetze:
On Grapheneos, current version

When you add a new User or do the first time initialisation of grapheneos, you are asked if you want to restore an existing backup. Choose to restore, plug in USB storage and you are asked for the 12 passphrase words. If you make a typo somewhere, simply no backups are found. There is only a red warning that it might be that the passphrase was wrong. 

There is no Back Button, which is the bug. You cannot check, re enter or correct your passphrase and try again. If you fail here, you have no chance to restore any backup. even not via systen/backup menu because of force to generate new passphrase. 

Only chance is to factory reset the whole phone or delete and create a new additional user profile.

grote:
The words should have a checksum, warning you about impossible combinations and unknown words. But yeah, we should allow retries in setup wizard.

Issue #739, opened 2024-09-05
Restore flow should allow returning to pre-storage selection


t-m-w:
Currently, if you attempt to restore a backup set, you are not able to change your mind and select a different backup set without wiping the data of Seedvault. This includes during the Setup Wizard: for example, if you start trying to restore a backup set and realize you chose the wrong one (e.g. inserted a USB drive with the wrong set), you cannot change your mind, and basically need to wipe the device if you still want to do your restore within the Setup Wizard.

grote:
related to https://github.com/seedvault-app/seedvault/issues/636

grote:
### Current situation

The restore and the backup flow currently use the _same_ saved key and the same saved backup location. This simplifies the code and allows restore after SUW easily. So we either allow the restore flow to _re-set_ key and location or we let the restore from SUW use a different _temporary_ location and key.

When Seedvault starts, be it for restore from SUW or for showing the normal UI, it checks if key and location have been set and if not, it prompts the user to provide what is missing. In both cases, it first asks for the location and then for the recovery code. In case of restore, before accepting the location, we check if there are even backups available there, but since we don't have the key, we can't check if we can also decrypt those backups.

So probably the most simple change would be to ask the user for the code first and then the location, so we can already ensure that the selected location contains backups we can decrypt. However, this doesn't help, if the user entered the wrong key which is something we saw happen in the wild already. So ideally, the location and the key can be re-set, if needed.

### How to allow resetting key and location?

One option would be to do the re-setting in SUW. Can SUW reset the app data of an app, e.g. `adb shell pm clear`?

Seedvault itself could clear the saved key or backup location. But when? After we can't find or can't decrypt backups in the restore flow? But only when started via SUW, not when the user entered the restore flow after a clean app start?

grote:
OK, I hadn't thought about the "change your mind" scenario, but rather when you entered actual wrong information and thus no backups were found. We'd have to think how to allow a restart still, maybe as a semi-hidden action in the screen that lists all the app backups.

PR #893, opened 2025-03-20
Allow try again or restart when no backups found during restore

### LINKED ISSUE #891: Restore only checks for app backups
1. I made a backup of my media files in Android 14 (matrixx os), but when i tried to restore them in A15 (same rom) I'm getting an error "we couldn't find any backups at this location" even tho there's clearly a backup (along with a .nomedia file and .snapshot file).

Image
Is this tool not backwards compatible or is it a bug?

grote:
Restart is only available during SUW as its potential to mess with the app is greater during normal operations.

Closes #891 #636

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/backend/BackendManager.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/backend/BackendManager.kt
@@ -102,6 +102,21 @@ class BackendManager(
     }
 
     /**
+     * Clears the storage plugins and current [BackendProperties].
+     *
+     * IMPORTANT: Do no call this while current plugins are being used,
+     *            e.g. while backup/restore operation is still running.
+     */
+    @WorkerThread
+    @Synchronized
+    fun removePlugins() {
+        settingsManager.clearStorageBackend()
+        mBackend = null
+        mBackendProperties = null
+        blobCache.clearLocalCache()
+    }
+
+    /**
      * Check if we are able to do backups now by examining possible pre-conditions
      * such as plugged-in flash drive or internet access.
      *
--- a/app/src/main/java/com/stevesoltys/seedvault/backend/saf/SafHandler.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/backend/saf/SafHandler.kt
@@ -53,12 +53,12 @@ internal class SafHandler(
     }
 
     /**
-     * Searches if there's really an app backup available in the given storage location.
+     * Searches if there's really an backup available in the given storage location.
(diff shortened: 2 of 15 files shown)


lucasmz-dev:
quick smoke test, seems to do what it describes

grote:
thanks a lot for testing lucas! Did you test the SUW restart flow as well?

lucasmz-dev:
I did, though I didn't actually try to check if inputting the correct key works and I probably should lmk if that helps

t-m-w:
> I did, though I didn't actually try to check if inputting the correct key works and I probably should lmk if that helps

@lucasmz-dev If you're willing, I'd greatly appreciate it!

t-m-w (changes requested):
If you enter the code incorrectly and start from the beginning, and now enter the right code, it still says the code is wrong. It's like it's not even trying to use the newly-entered code, no matter how many times you tap Try Again or start from the beginning. I needed to `adb shell pm clear com.stevesoltys.seedvault` to get it to accept the new code, so this isn't working yet.

grote:
I pushed a change that will kill the app to get the new key material. Couldn't test it with SUW though.

t-m-w (approved):
Approved because it's an overall big improvement over status quo and it works now.

About the issues it would close:

Should close #636 for sure.

#739, not quite. Although this solves the situation where you entered the wrong recovery code or where the storage medium has become unavailable, it only allows you to select a different underlying storage under those failure conditions, not simply because you feel like it or you made a mistake. For example, if you realize you used the WebDAV info for the wrong device, it seems like this wouldn't allow you to undo that. What you *could* probably due is go into airplane mode or go back and turn off Wi-Fi and mobile data to induce a failure. And if it's a USB drive, you can remove it to induce a failure. This does provide a workaround, but not a solution. (edit: Let me know if I'm just missing something...)

Not fully tested, but *should* help for #891 as well, since you can skip restoring apps if none are found.

grote:
Thanks for testing again!

Ok, so I removed #739 from the close list. 

After you pressed the restart button, what happened? The app closed itself and brought you back to SUW where you could launch it again normally and it then accepted the correct recovery code?

t-m-w:
> After you pressed the restart button, what happened? The app closed itself and brought you back to SUW where you could launch it again normally and it then accepted the correct recovery code?

That's right, just as you described. It looked seamless to me, since the button does say "Start from the beginning", and then you find yourself back at that SUW page asking if you want to restore or not, choose to do so, and smooth sailing from there.

grote:
Nice, thanks for clarifying.

PR #901, opened 2025-03-31
Treat nested EOFException as transient

grote:


t-m-w (approved):
This time, unlike usual, I've looked at the code but have not tested anything.

It looks good to me.

grote:
thanks for the review!

PR #902, opened 2025-04-01
Add progress bar to app backup pruning notification

grote:


t-m-w (approved):
I see an indeterminate progress bar briefly which transforms into a stunning actual-progress bar and speed-runs from an early step to the end so fast I often can't see it. On cuttlefish, though, the steps are more pronounced. Either way, from logs, I know it's succeeding. Long story short, this looks like it works, since it makes it to pruning 21/21 or whatever the number is. Seems good to me.

grote:
You'll be able to enjoy it for longer when there's more data to prune ;) Thanks for testing!

PR #916, opened 2025-05-09
Try to show app backup notification in DND mode

### LINKED ISSUE #913: App backup notification does not bypass DND
This caused me some confusion as I thought the backup process wasn't going working while testing and I was just ejecting and redoing it being like 'WHAT THE HELL' because I didn't really think of DND + the initial backup file notifications worked fine.

# Steps to reproduce
1. Configure all filtered notifications settings to be hidden in the 'suspended tab' (this is PT, it's the last setting)
2. Trigger a backup 
3. Observe progress notifications in notifications tab

You'll notice the backup progress notification for apps does not show up, files does, and cleanup seems to as well. Finished backup notifications don't go through but that might be expected, not sure what the standard is for that... What I wouldn't say is expected are the progress related ones.

grote:
Closes #913


grote:
@lucasmz-dev maybe you can test this?

lucasmz-dev:
I can't :(

grote:
@t-m-w was so nice to test, but doesn't seem this helps.

PR #924, opened 2025-05-29
Fix hardcoded "Remove" string

lucasmz-dev:
Found in the vertical ... menu found on a user-added item in storage backups


chirayudesai (approved):
Thanks!

PR #927, opened 2025-05-31
Try to follow M3 guidelines for text

lucasmz-dev:
Not really planning this in a formal way just kind of looking through it and being like 'I don't like that, let me change it'

In this I did notice that some strings mention 'Automatic restore' as the only solution or "What you need to do" but now we have the regular restore so...

I wanna read some more about Google's M3 rules for text as well but I haven't yet

Consider this a 'draft' but also not if you find this a good thing?

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -15,23 +15,23 @@
 
     <!-- Settings -->
     <string name="settings_category_apps">App backup</string>
-    <string name="settings_backup">Backup my apps</string>
+    <string name="settings_backup">Back up apps</string>
     <string name="settings_backup_location">Backup location</string>
     <string name="settings_backup_location_none">None</string>
     <string name="settings_backup_location_internal">Internal storage</string>
     <string name="settings_backup_last_backup_never">Never</string>
-    <string name="settings_backup_dialog_title">Really turn off app backups?</string>
-    <string name="settings_backup_dialog_message">When you turn backups on again, the backup process may take longer than usual, and it will use additional storage space.</string>
+    <string name="settings_backup_dialog_title">Turn off app backups?</string>
+    <string name="settings_backup_dialog_message">When you turn backups on again, the backup process may take longer than usual, and it\'ll use additional storage space.</string>
     <string name="settings_backup_dialog_disable">Turn off anyway</string>
-    <string name="settings_info">All backups are encrypted on your phone. To restore from backup you will need your 12-word recovery code.</string>
+    <string name="settings_info">All backups are encrypted on your phone. To restore from backup you\'ll need your 12-word recovery code.</string>
     <string name="settings_auto_restore_title">Automatic restore</string>
-    <string name="settings_auto_restore_summary">When reinstalling an app, restore backed up settings and data.</string>
-    <string name="settings_auto_restore_summary_usb">Note: Your %1$s needs to be plugged in for this to work.</string>
+    <string name="settings_auto_restore_summary">When reinstalling an app, backed up data and settings will be restored automatically.</string>
+    <string name="settings_auto_restore_summary_usb">Note: Your "%1$s" device needs to be plugged in for this to work.</string>
     <string name="settings_category_app_data_backup">App data backup</string>
     <string name="settings_backup_apk_title">APK backups</string>
-    <string name="settings_backup_apk_summary">Backs up the apps themselves. Otherwise, only app data will get backed up.</string>
-    <string name="settings_backup_apk_dialog_title">Really disable APK backups?</string>
-    <string name="settings_backup_apk_dialog_message">After disabling APK backups, app data will still be backed up. However, it will not get restored automatically.\n\nYou will need to install all your apps manually while having \"Automatic Restore\" switched on.</string>
+    <string name="settings_backup_apk_summary">Backs up the apps themselves. If disabled, only app data will be backed up.</string>
+    <string name="settings_backup_apk_dialog_title">Disable APK backups?</string>
+    <string name="settings_backup_apk_dialog_message">After disabling APK backups, app data will still be backed up. However, you\'ll need to restore manually.</string>
     <string name="settings_backup_apk_dialog_cancel">Cancel</string>
     <string name="settings_backup_apk_dialog_disable">Disable APK backups</string>
     <string name="settings_backup_status_title">Backup status</string>
@@ -42,23 +42,23 @@
     <string name="settings_backup_status_next_backup_usb">Backups will happen automatically when you plug in your USB drive</string>
     <string name="settings_backup_scheduling_title">Backup scheduling</string>
     <string name="settings_backup_app_check_title">Check integrity</string>
-    <string name="settings_backup_app_check_summary">Ensure that backup is working for restore</string>
+    <string name="settings_backup_app_check_summary">Make sure the backup isn\'t corrupted</string>
     <string name="settings_backup_exclude_apps">Exclude apps</string>
-    <string name="settings_backup_now">Backup now</string>
+    <string name="settings_backup_now">Back up now</string>
     <string name="settings_category_storage">File backup (beta)</string>
-    <string name="settings_backup_storage_title">Backup my files</string>
+    <string name="settings_backup_storage_title">Back up files</string>
     <string name="settings_backup_files_title">Included files and folders</string>
     <string name="settings_backup_files_summary">None</string>
-    <string name="settings_backup_file_check_title">Check backup integrity</string>
-    <string name="settings_backup_file_check_summary">Make sure my backup is working for restoring files</string>
+    <string name="settings_backup_file_check_title">Check integrity</string>
+    <string name="settings_backup_file_check_summary">Make sure the file backup isn\'t corrupted</string>
     <string name="settings_backup_recovery_code">Recovery code</string>
-    <string name="settings_backup_recovery_code_summary">Verify existing code or generate a new one</string>
+    <string name="settings_backup_recovery_code_summary">Verify the existing code or generate a new one</string>
     <string name="settings_backup_storage_dialog_title">Beta feature</string>
-    <string name="settings_backup_storage_dialog_message">Backing up files is beta and might not work. Do not rely on it for important data.</string>
+    <string name="settings_backup_storage_dialog_message">File backup is still in beta and might not work. Don\'t rely on it for important data.</string>
     <string name="settings_backup_storage_dialog_ok">Enable anyway</string>
     <string name="settings_backup_storage_battery_optimization">Warning: No automatic backups, because battery optimization is active.</string>
     <string name="settings_backup_new_code_dialog_title">New recovery code required</string>
-    <string name="settings_backup_new_code_dialog_message">To continue using app backups, you need to generate a new recovery code.\n\nWe are sorry for the inconvenience.</string>
+    <string name="settings_backup_new_code_dialog_message">To continue using app backups, you need to generate a new recovery code.\n\nSorry for the inconvenience.</string>
     <string name="settings_backup_new_code_code_dialog_ok">New code</string>
     <string name="settings_backup_metered_title">Backup stopped</string>
     <string name="settings_backup_metered_text">The backup won\'t proceed because your device is using mobile data.\n\nYou can enable backups on mobile data under \"Backup scheduling\".</string>
@@ -74,18 +74,18 @@
     <string name="settings_scheduling_charging_title">Back up only when charging</string>
 
     <string name="settings_app_check_title">Verify app backup integrity</string>
-    <string name="settings_app_check_text">Each backup process will automatically check the backup integrity. However, to save time, actual app data will not be downloaded and verified.</string>
-    <string name="settings_app_check_text2">Here you can run a one-time check to verify your app backup. Select how much of the app data should be downloaded for the check. The more you select, the longer it will take, yet the more reliable the check will be. This will run in the background and show a notification once the check is done.</string>
+    <string name="settings_app_check_text">When a backup runs, it\'ll do a structural check of the backup. However, this doesn\'t include the actual backup data.</string>
+    <string name="settings_app_check_text2">Here you can run a one-time check to verify your app backup. Select how much of the app data should be downloaded for the check. The more you select, the longer it\'ll take, yet the more reliable the check will be. This will run in the background and show a notification once the check is done.</string>
     <string name="settings_app_check_warning">Warning: Downloading large amounts of data can take a long time.</string>
     <string name="settings_app_check_button">Check backup</string>
     <string name="settings_file_check_title">Verify file backup integrity</string>
     <string name="settings_file_check_text">Here you can run a one-time check to verify your file backup.</string>
-    <string name="settings_file_check_text2">Select how much of file data should be downloaded for the check. The more you select, the longer it will take, yet the more reliable the check will be.\n\nThe check will run in the background and send you a notification once it is done.</string>
+    <string name="settings_file_check_text2">Select how much of file data should be downloaded for the check. The more you select, the longer it\'ll take, yet the more reliable the check will be.\n\nThe check will run in the background and send you a notification once it is done.</string>
 
     <string name="settings_expert_title">Expert settings</string>
     <string name="settings_expert_logcat_title">Save app log</string>
     <string name="settings_expert_logcat_summary">Developers can diagnose bugs with these logs.\n\nWarning: The log file might contain personally identifiable information. Review before and delete after sharing!</string>
-    <string name="settings_expert_logcat_error">Error: Could not save app log</string>
+    <string name="settings_expert_logcat_error">Error: Couldn\'t save app log</string>
 
     <!-- Storage Location -->
     <string name="storage_fragment_backup_title">Choose where to store backups</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -42,26 +42,26 @@
     <string name="settings_backup_status_next_backup_usb">Backups will happen automatically when you plug in your USB drive</string>
     <string name="settings_backup_scheduling_title">Backup scheduling</string>
     <string name="settings_backup_app_check_title">Check integrity</string>
-    <string name="settings_backup_app_check_summary">Ensure that backup is working for restore</string>
+    <string name="settings_backup_app_check_summary">Ensure that backup\'s working for restore</string>
```
lucasmz-dev: TODO: *the* backup

app/src/main/res/values/strings.xml
```diff
@@ -42,26 +42,26 @@
...
     <string name="settings_backup_now">Backup now</string>
     <string name="settings_category_storage">File backup (beta)</string>
-    <string name="settings_backup_storage_title">Backup my files</string>
+    <string name="settings_backup_storage_title">Backup files</string>
     <string name="settings_backup_files_title">Included files and folders</string>
     <string name="settings_backup_files_summary">None</string>
-    <string name="settings_backup_file_check_title">Check backup integrity</string>
-    <string name="settings_backup_file_check_summary">Make sure my backup is working for restoring files</string>
+    <string name="settings_backup_file_check_title">Check integrity</string>
+    <string name="settings_backup_file_check_summary">Ensure that file backup\'s working for restore</string>
```
lucasmz-dev: the backup, again

app/src/main/res/values/strings.xml
grote: is there an official recommendation to use `backup's working` instead of `backup is working`? Is this even proper English?

gdt: With native en_US hat on, it should remain "backup is working".   The use of `'` can be confusing, being both for possessive, and for contractions.   A contraction example is "isn't".    A [confusing example](https://grammarist.com/spelling/its-its/) is "it's" for "it is", and "its" for the possessive  My perception is that as general rule, contractions should be avoided in formal writing, but opinions differ.   Personally, I think it is better to avoid contractions in text that will be read by non-native speakers on the grounds that they are not helpful and increase the required reading skill level.   We don't really have a "strings should be in simplified English when possible" norm, but I think a bias in that direction is good, as most people are not native Engilsh speakers but feel they have to learn some to cope with computers, as English is now the *lingua franca*.

I don't believe it is proper to form contractions "foo's" from random "foo is".  I found a [page with 80 standard contractions](https://editorsmanual.com/articles/contractions-list/).  Surprising to me, it says that random nouns and is can become a contraction.   However, I stand by the idea that it is unhelpful to the reader and too informal.

Stepping back, I don't find "backup is working for restore" to be clear.

app/src/main/res/values/strings.xml
```diff
@@ -20,18 +20,18 @@
     <string name="settings_backup_location_none">None</string>
     <string name="settings_backup_location_internal">Internal storage</string>
     <string name="settings_backup_last_backup_never">Never</string>
-    <string name="settings_backup_dialog_title">Really turn off app backups?</string>
+    <string name="settings_backup_dialog_title">Are you sure you want to turn off app backups?</string>
```
gdt: hard to say if this is an improvement.   more formal, but longer.

app/src/main/res/values/strings.xml
```diff
@@ -220,7 +220,7 @@
     <!-- App Backup and Restore State -->
 
     <string name="backup_section_system">System data</string>
-    <string name="backup_sms">SMS text messages</string>
+    <string name="backup_sms">SMS messages</string>
```
gdt, line 223: This is probably good, as so many people use "text" with varying meanings.

app/src/main/res/values/strings.xml
```diff
@@ -229,18 +229,18 @@
     <string name="backup_app_stopped_button">Launch</string>
     <string name="backup_app_stopped_no_intent">Cannot launch app</string>
     <!-- This text gets shown for apps that the OS did not try to backup for whatever reason e.g. no backup was run yet -->
-    <string name="backup_app_not_yet_backed_up">Waiting to back up…</string>
+    <string name="backup_app_not_yet_backed_up">Awaiting backup…</string>
```
gdt, line 232: I think the original wording is better.

app/src/main/res/values/strings.xml
```diff
@@ -229,18 +229,18 @@
...
     <string name="restore_app_not_yet_backed_up">Was not yet backed up</string>
     <string name="backup_app_in_progress">Backing up</string>
     <string name="backup_app_success">Backed up</string>
     <string name="backup_app_warning">Backup warning</string>
-    <string name="backup_app_was_stopped">Not backed up as it wasn\'t used recently</string>
-    <string name="restore_app_was_stopped">Was not backed up as it hadn\'t been used recently</string>
+    <string name="backup_app_was_stopped">Not backed up as it is force-stopped</string>
+    <string name="restore_app_was_stopped">Was not backed up as it was force-stopped</string>
```
gdt: It is not clear that this change is correct.   Say what is really happening.  Perhaps "Not backed up because app is not running."

app/src/main/res/values/strings.xml
```diff
@@ -165,7 +165,7 @@
     <string name="notification_init_text">Preparing existing backup data for reuse…</string>
     <string name="notification_finalizing_text">Finalizing backup…</string>
     <string name="notification_apk_text">Backing up APK of %s</string>
-    <string name="notification_apk_not_backed_up">Saving list of apps we can not back up.</string>
+    <string name="notification_apk_not_backed_up">Saving a list of apps that can\'t be backed up</string>
```
gdt, line 168: can not, not "can't".

app/src/main/res/values/strings.xml
```diff
@@ -181,13 +181,13 @@
     <string name="notification_error_action">Fix</string>
 
     <string name="notification_space_error_title">Insufficient backup space</string>
-    <string name="notification_space_error_text">Your backup location is running out of space. Free up some space so backups can run.</string>
+    <string name="notification_space_error_text">Your backup location\'s running out of space. Free up some space so backups can run.</string>
```
gdt: objection :-)

app/src/main/res/values/strings.xml
```diff
@@ -181,13 +181,13 @@
...
 
     <string name="notification_restore_channel_title">Restore notification</string>
     <string name="notification_restore_title">Restore running</string>
 
     <string name="notification_restore_error_channel_title">Auto restore flash drive error</string>
-    <string name="notification_restore_error_title">Could not restore data for %1$s</string>
+    <string name="notification_restore_error_title">Couldn\'t restore data for %1$s</string>
```
gdt, line 190: In general, I am opposed to all changes toward contractions.

app/src/main/res/values/strings.xml
```diff
@@ -203,19 +203,19 @@
...
+    <string name="backup_app_check_success_disclaimer">Note: Can\'t verify whether individual apps include all of their user data in the backup, or if backups were tested by the developer.</string>
     <string name="backup_app_check_error_title" translatable="false">@string/notification_checking_error_title</string>
-    <string name="backup_app_check_error_no_snapshots">We could not find any backup. Please run a successful backup first and then try checking again.</string>
-    <string name="backup_app_check_error_only_broken_snapshots">We found %1$d backup snapshots. However, all of them were corrupted. Please run a successful backup and then try checking again.</string>
-    <string name="backup_app_check_error_has_snapshots">We found %1$d backup snapshots, some of them are corrupt or have errors. Below are the backups that could be restored.</string>
-    <string name="backup_app_check_error_no_result">We lost the detailed results. Did a long time pass since running the check? Try running again.</string>
+    <string name="backup_app_check_error_no_snapshots">Couldn\'t find any backup. Please run a successful backup first and then try checking again.</string>
```
gdt, line 215: It is good to get rid of "we".

app/src/main/res/values/strings.xml
```diff
@@ -268,32 +268,32 @@
...
-    <string name="restore_storage_in_progress_info">Your files are being restored in the background. You can use your phone while this is running.\n\nSome apps (e.g. Signal or WhatsApp) might require files to be fully restored to import a backup. Try to avoid starting those apps before file restore is complete.</string>
+    <string name="restore_storage_in_progress_info">Your files are being restored in the background. You can use your phone while this is running.\n\nSome apps, like Signal or WhatsApp, might require files to be fully restored to import a backup. Try to avoid starting those apps before file restore\'s complete.</string>
     <string name="restore_storage_got_it">Got it</string>
 
     <string name="storage_internal_warning_title">Warning</string>
-    <string name="storage_internal_warning_message">You have chosen internal storage for your backup. This will not be available when your phone is lost or broken.</string>
+    <string name="storage_internal_warning_message">You\'ve chosen internal storage for your backup. This won\'t be available when your phone\'s lost or broken.</string>
```
gdt: random noun contraction

app/src/main/res/values/strings.xml
```diff
@@ -42,26 +42,26 @@
...
-    <string name="settings_backup_file_check_title">Check backup integrity</string>
-    <string name="settings_backup_file_check_summary">Make sure my backup is working for restoring files</string>
+    <string name="settings_backup_file_check_title">Check integrity</string>
+    <string name="settings_backup_file_check_summary">Ensure that file backup\'s working for restore</string>
     <string name="settings_backup_recovery_code">Recovery code</string>
-    <string name="settings_backup_recovery_code_summary">Verify existing code or generate a new one</string>
+    <string name="settings_backup_recovery_code_summary">Verify the existing code or generate a new one</string>
     <string name="settings_backup_storage_dialog_title">Beta feature</string>
-    <string name="settings_backup_storage_dialog_message">Backing up files is beta and might not work. Do not rely on it for important data.</string>
+    <string name="settings_backup_storage_dialog_message">File backup\'s still in beta and might not work. Don\'t rely on it for important data.</string>
```
stevesoltys: "Support for backing up files is a beta feature. Don't rely on this for important data."

Perhaps?

app/src/main/res/values/strings.xml
```diff
@@ -42,26 +42,26 @@
...
     <string name="settings_backup_storage_dialog_ok">Enable anyway</string>
-    <string name="settings_backup_storage_battery_optimization">Warning: No automatic backups, because battery optimization is active.</string>
+    <string name="settings_backup_storage_battery_optimization">Warning: No automatic backups, because battery optimization\'s active.</string>
```
stevesoltys: I personally prefer how this one was.

app/src/main/res/values/strings.xml
```diff
@@ -42,26 +42,26 @@
...
     <string name="settings_backup_new_code_dialog_title">New recovery code required</string>
-    <string name="settings_backup_new_code_dialog_message">To continue using app backups, you need to generate a new recovery code.\n\nWe are sorry for the inconvenience.</string>
+    <string name="settings_backup_new_code_dialog_message">To continue using app backups, you need to generate a new recovery code.\n\nSorry for the inconvenience.</string>
     <string name="settings_backup_new_code_code_dialog_ok">New code</string>
     <string name="settings_backup_metered_title">Backup stopped</string>
-    <string name="settings_backup_metered_text">The backup won\'t proceed because your device is using mobile data.\n\nYou can enable backups on mobile data under \"Backup scheduling\".</string>
+    <string name="settings_backup_metered_text">The backup won\'t proceed because your device\'s using mobile data.\n\nYou can enable backups on mobile data under \"Backup scheduling\".</string>
```
stevesoltys: Personally prefer the before here.

app/src/main/res/values/strings.xml
```diff
@@ -99,7 +99,7 @@
     <string name="storage_fake_nextcloud_title" translatable="false">Nextcloud</string>
     <string name="storage_fake_nextcloud_summary">Tap to install</string>
     <string name="storage_fake_nextcloud_summary_installed">Tap to set up account</string>
-    <string name="storage_fake_nextcloud_summary_unavailable">Account not available. Set one up (or disable passcode).</string>
+    <string name="storage_fake_nextcloud_summary_unavailable">Account not available. Set one up, or disable the app\'s passcode.</string>
```
stevesoltys, line 102: Or disable your passcode in Nextcloud? Not sure the context here

app/src/main/res/values/strings.xml
```diff
@@ -143,18 +143,18 @@
     <string name="recovery_code_input_hint_10">Word 10</string>
     <string name="recovery_code_input_hint_11">Word 11</string>
     <string name="recovery_code_input_hint_12">Word 12</string>
-    <string name="recovery_code_error_empty_word">You forgot to enter this word.</string>
-    <string name="recovery_code_error_invalid_word">Wrong word.</string>
-    <string name="recovery_code_error_checksum_word">Your code is invalid. Please check all words as well as their position and try again!</string>
+    <string name="recovery_code_error_empty_word">You forgot to enter this word</string>
+    <string name="recovery_code_error_invalid_word">Wrong word</string>
+    <string name="recovery_code_error_checksum_word">Your code\'s invalid. Please check all words as well as their position and try again!</string>
```
stevesoltys: Prefer the former personally

app/src/main/res/values/strings.xml
```diff
@@ -143,18 +143,18 @@
...
     <string name="recovery_code_verification_ok_title">Recovery code verified</string>
-    <string name="recovery_code_verification_ok_message">Your code is correct and will work for restoring your backup.</string>
+    <string name="recovery_code_verification_ok_message">Your code\'s correct and will work when you need to restore.</string>
```
stevesoltys: Anywhere that we're just replacing is with apostrophe s feels off to me (personally).

app/src/main/res/values/strings.xml
lucasmz-dev: Oops. Not intended. This should be "Turn off app backups?"

app/src/main/res/values/strings.xml
lucasmz-dev: Isn't the definition of running in Android kind of all over the place? It's not really clear to the user what each thing means, it may not be actively running or not, and it can still be backed up, no?

I feel like oversimplifying might be hurtful here

gdt: 1. I thought that the OS could stop an app (that was in background not doing anything) if the OS believes it isn't being used.   I have the further impression this is no different from force stopped, from seedvault point of view.  (PS there I could using contractions, but I'm just commenting, not writing docs, and "isn't' is one of the most common usages.
2. In general I believe that the Android app practice that there is not documentation is regrettable.    This discussion points out that a longer explanation is needed in the manual, and maybe it's there, which raises the question of how to point to it.

app/src/main/res/values/strings.xml
lucasmz-dev: I think I overdone contractions. I'm not a native English speaker (is anyone here?) and I didn't really learn all the formality and stuff, literally everything was from informal conversation. 

grote isn't either AFAIK, not sure about Steve. So it could end up being literally a bunch of non natives trying to figure out what to do. Quite common in FOSS tbh 😅

I think I'll go ahead and fix contractions to not be used for nouns.

For reference, this was from https://m3.material.io/foundations/content-design/style-guide/grammar-and-punctuation

app/src/main/res/values/strings.xml
lucasmz-dev: Yeah I'm not a fan of that. I feel like it makes the software too personal to the developer

app/src/main/res/values/strings.xml
gdt: I did not know about the M3 style guide.  Given that, I really can't object to the normal contractions, the top 10 or so that aren't very frequent.   I am still concerned that non-native speakers trying to read English will have a harder time.

The style guide does not suggest noun's for "noun is", by my reading.  Plus, I find it very confusing, as a native speaker, and think it would be even harder to parse for a non-native speaker.

gdt: There's also that en_GB rules are slightly different.   I don't know if Android writes in en_US and en_GB, en_AU, en_NZ, en_CA get lightly translated.  My friends Bruce and Sheila would want an AU-specific translation.

app/src/main/res/values/strings.xml
lucasmz-dev: This line is a bit confusing to me overall. What does this even mean?
From the string name, I would guess it would mean like 'oh like system apps that don't need to have APKs backed up' but then from the text it then sounds like APKs that break things, aren't allowed to be backed up
If you think from the name, it could also sound like it is saying "We can >NOT BACK UP<" as in you don't need to back up

app/src/main/res/values/strings.xml
lucasmz-dev: Yeah need to check

Isn't Nextcloud backup, like... disabled anyway right now though? Is this string even possible to reach

app/src/main/res/values/strings.xml
lucasmz-dev: There are variants of English in Weblate, but they're not well maintained; Australia and United Kingdom are the ones that even have translations. I had to manually fix them a while ago for source-string changes otherwise they'd still be using older variants, they were literally the same so...
I don't know if anything even changes between them, honestly

app/src/main/res/values/strings.xml
lucasmz-dev: phone no longer has a contraction

app/src/main/res/values/strings.xml
lucasmz-dev: I don't really agree for this case, but what about "It wasn't possible to" ?

PR #930, opened 2025-06-05
Don't request backup more than once when we are notified about available USB storage

### LINKED ISSUE #923: Repeated backup processes running at once
Image

Happened two times now, one day and then the other, I got this issue where after connecting my USB, it triggered backups but then created two backup processes. e.g. there were two processes like files and APK backup happening together. This then broke them and they fail at the end, they restart once and then die at the end. 

Logs sent privately to grote

grote:
Test instructions:
* set up backup to usb
* eject drive
* fake system time or code
* insert drive
* observe/save logcat and system notification that it actually starts a new backup

You'll see `UsbIntentReceiver` logging `Last backup older than it should be, requesting a backup...` or `We have a recent backup, not requesting a new one.`.

You can't really test that this fixes the multiple start issue, but what is important is that it still does backups at all ;)

Starting a backup manually would also be nice to test.

Extra hard test scenario: Insert and remove and re-insert flashdrive short after. Ideally, we can detect that it had already requested a backup and won't request a second one.

May close #923

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/UsbMonitorService.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/UsbMonitorService.kt
@@ -20,6 +20,7 @@ import com.stevesoltys.seedvault.ui.notification.BackupNotificationManager
 import com.stevesoltys.seedvault.ui.notification.NOTIFICATION_ID_USB_MONITOR
 import com.stevesoltys.seedvault.worker.BackupRequester.Companion.requestFilesAndAppBackup
 import org.koin.android.ext.android.inject
+import java.util.concurrent.atomic.AtomicBoolean
 
 private const val TAG = "UsbMonitorService"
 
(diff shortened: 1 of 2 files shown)


grote:
@lucasmz-dev do you have a phone to test this?

app/src/main/java/com/stevesoltys/seedvault/UsbMonitorService.kt
```diff
@@ -39,12 +40,18 @@ class UsbMonitorService : Service() {
         val rootsUri = intent.data ?: error("No URI in start intent.")
         val contentResolver = contentResolver
         val handler = Handler(Looper.myLooper() ?: error("no looper"))
+        val hasRequestedBackup = AtomicBoolean(false)
```
t-m-w, line 43: Should this ever get reset back to `false`?

lucasmz-dev:
unfortunately not unless it was like calyxos signed

app/src/main/java/com/stevesoltys/seedvault/UsbMonitorService.kt
grote: Great question! The answer is `no`, because each time, we recognize that the right flash drive was inserted, we start a new `UsbMonitorService` where `onStartCommand` gets called. The `hasRequestedBackup` is local to that method, so we use it only once:

While it is still `false` we request backup and as soon as it became true we never request another backup with this observer. We are actually removing it anyway, because we want this specific observer to only trigger once.

t-m-w (approved):
This looks good to me. Logcat results look normal. I do see an instance where `onChange() requesting backup now!` happens five times within the same 50ms, but it doesn't lead to multiple backups.

I tried the extra hard test, but it leads to me needing to format the drive each time, so "extra hard" is right. One thing I noticed, however, is that when I disconnect the drive, it notifies me that the backup failed and suggests connecting to Wi-Fi... But that's off-topic for this pull request.

t-m-w:
> This looks good to me. Logcat results look normal. I do see an instance where `onChange() requesting backup now!` happens five times within the same 50ms, but it doesn't lead to multiple backups.
> 
> I tried the extra hard test, but it leads to me needing to format the drive each time, so "extra hard" is right. One thing I noticed, however, is that when I disconnect the drive, it notifies me that the backup failed and suggests connecting to Wi-Fi... But that's off-topic for this pull request.

I managed to get these results *without actually having this change* pulled in. So I will have to try again.

t-m-w (approved):
So, what would be really useful is if I could get the same result as yesterday, where I had "onChange() requesting backup now!" show up multiple times in a row, because then I could use the other *new* log entries as a gauge of whether things are working correctly. But it's only showing up once.

We have a funny situation: yesterday, it seemed I was observing symptoms that could lead to the problem, but no problem, without patch. Today, with patch, the symptoms are gone, too. Very difficult to confirm that this helps if I can't confirm anything was wrong to begin with.

The best I can think to do is see if it hurts.

So far, with USB, everything seems fine, including manual backups.

I tried backing up and restoring with WebDAV and that worked fine, too.

PR #932, opened 2025-06-06
Use more reliable detection of flash drive presence

### LINKED ISSUE #931: Auto-restore or single app restore with flash-drive failing on Android 15
In work profile, when uninstalling an app and then restoring it with Seedvault, it fails.

It also sends a notification about needing to plug in my storage device... which is plugged in and in the restore process anyway 🤷

Image
Image

OS: CalyxOS 6.7.23 (Android 15 QPR2)
Device: rhode "moto g52"
Seedvault version: 15-5.6
Backup location: USB Drive

grote:
Test instructions: back up to flash drive. uninstall app, re-install app. see if auto-restore works. try manually restoring a single app.

Fixes #931

### DIFF
--- a/core/src/main/java/org/calyxos/seedvault/core/backends/saf/SafProperties.kt
+++ b/core/src/main/java/org/calyxos/seedvault/core/backends/saf/SafProperties.kt
@@ -7,8 +7,11 @@ package org.calyxos.seedvault.core.backends.saf
 
 import android.content.Context
 import android.net.Uri
+import android.provider.DocumentsContract
 import android.provider.DocumentsContract.Root.COLUMN_ROOT_ID
+import android.util.Log
 import androidx.annotation.WorkerThread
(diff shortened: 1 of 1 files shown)


t-m-w:
Manually restore is working. Auto-restore is not working.

Tested with Fossify Notes.

t-m-w (changes requested):
If we want auto-restore to work as part of this change, looks like more work is needed?

t-m-w (approved):
Never mind, this is working even with auto-restore. I was going to provide logs on the differences, and I noticed that for manual restore, a different restore set was being used. I think there were side effects caused by my "time traveling" when testing another PR...? I set the clock both forward and back in time. Not recommended.

grote:
Yeah auto-restore failure is likely due to messing with clock as this PR doesn't mess with this.

Issue #864, opened 2025-02-06
android14 branch not working in emulator


projectgus:
I've been trying to get the emulator to run according to the instructions in `app/development/DEVELOPMENT.md` and the android14[*] branch (commit 1229ceaf) and Android Studio 2024.2.2, but I haven't had much luck:

Running the gradle `app:provisionEmulator` task appears to complete successfully ([log output](https://gist.github.com/projectgus/188731bf288dc5f8bb4baa0d167f993f/raw/27c24d7d0e4360b63b28158bd1626030d3fc67e5/01-app:provisionEmulator.log)). I can see the emulator window at this point.

Running the runtime `app-emulator` config launches `app:installEmulatorRelease` and this seems to complete OK ([log output](https://gist.github.com/projectgus/188731bf288dc5f8bb4baa0d167f993f/raw/27c24d7d0e4360b63b28158bd1626030d3fc67e5/02-app:InstallEmulatorRelease.log)).

Then Android Studio switches tabs and shows the run output:

```
2025-02-06 16:16:05: Launching app-emulator on 'seedvault'.
Open logcat panel for emulator seedvault
```

[Default logcat panel output (filtered)](https://gist.github.com/projectgus/188731bf288dc5f8bb4baa0d167f993f/raw/27c24d7d0e4360b63b28158bd1626030d3fc67e5/03-logcat-filtered.log).

[Same logcat without filter](https://gist.github.com/projectgus/188731bf288dc5f8bb4baa0d167f993f/raw/27c24d7d0e4360b63b28158bd1626030d3fc67e5/04-logcat-unfiltered.log).

The emulator doesn't appear to do anything in response. If I manually open Settings -> System -> Backup then Seedvault doesn't seem to load at all:

![Image](https://github.com/user-attachments/assets/3dd61f70-647a-4d71-848b-08a8d6901817)

[Logcat for opening the Backup menu item](https://gist.github.com/projectgus/188731bf288dc5f8bb4baa0d167f993f/raw/27c24d7d0e4360b63b28158bd1626030d3fc67e5/05-logcat-open-backup.log).

[All log files from this issue report can be downloaded from this gist](https://gist.github.com/projectgus/188731bf288dc5f8bb4baa0d167f993f).

When I have time I'll try again from the android13 branch.

Thanks for all of your effort developing and maintaining SeedVault!

([*] I had even less luck with the android15 branch, but it seems like that one probably isn't expected to work yet. Can provide details if useful.)

[...]

PR #960, opened 2025-10-11
[A15] Fix android studio emulator scripts

baschi29:
As discussed in #864.

I also slightly updates the scripts descriptions in `DEVELOPMENT.md`.

### DIFF
--- a/app/development/scripts/provision_emulator.sh
+++ b/app/development/scripts/provision_emulator.sh
@@ -59,7 +59,7 @@ fi
 ADB="$ANDROID_HOME/platform-tools/adb -s $EMULATOR_DEVICE_NAME"
 
 echo "Waiting for emulator to boot..."
-$ADB wait-for-device shell "while [[ -z $(getprop sys.boot_completed) ]]; do sleep 1; done;"
+$ADB wait-for-device shell "while [[ -z \$(getprop sys.boot_completed) ]]; do sleep 1; done;"
 
 echo "Provisioning emulator for write access to '/system'..."
 $ADB root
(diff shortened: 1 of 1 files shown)


baschi29:
@stevesoltys I updated my PR as discussed in the corresponding issue:

- [x] Escape using a backslash to maintain consistency of using double quotes throughout the scripts.
- [x] Gradle tasks and other scripts left untouched

I did not update the documentation, but choose to add a method to reliably set the transport directly in the script - this should be the intuitive way for people outside of the project and setting the transport again later on should be possible without errors. I will add a comment to the corresponding parts in the script. 

If this looks good to you, I will add the changes to the other branches and update the PRs accordingly :)

app/development/scripts/provision_emulator.sh
```diff
@@ -100,4 +100,10 @@ $ADB shell rm /sdcard/backup.tar.gz
 # sometimes a system dialog (e.g. launcher stopped) is showing and taking focus
 $ADB shell am broadcast -a android.intent.action.CLOSE_SYSTEM_DIALOGS
 
+echo "Enabling bmgr and setting transport"
+$ADB shell bmgr enable true
+# wait for the transport to be recognized by android
+adb shell "while [[ -z \$(bmgr list transports | grep seedvault) ]]; do sleep 1; done;"
```
baschi29, line 106: I noticed that the emulator just needs some time to pick up the new transport.
For some time related to the speed of the emulation it will not appear when calling `bmgr list transports`. After it appears, setting it can be done without errors, so the code just waits for this to happen.

It seems this could possibly also occur in the cirrus pipeline, but apparently the emulation there is simply fast enough so that this problem never appears.

stevesoltys (approved):

PR #1012, opened 2026-05-17
docs: add Seednaut utility to third-party tools

Baltram:
Hi! I recently released Seednaut, an open-source CLI utility for inspecting/extracting Seedvault backups.

It supports current v2 app backups as well as v0 file backups.

Since the README already lists third-party tooling for older backup formats, I thought it could be useful to include Seednaut as an option for users working with current backups too.

Happy to adjust the wording if preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,17 +87,17 @@ Seedvault can be translated with Weblate, as provided by [CalyxOS](https://hoste
 > The Seedvault developers make no guarantees about external software projects.
 > Please be aware that disclosing your secret recovery key to other software has security risks.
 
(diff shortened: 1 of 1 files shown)


grote:
Great work! Changes look good to me. However, what do you think about re-ordering the text so that tools handling newest format are listed first and legacy tools for older formats came last: v2, v1, v0.

As file backups only have one version, maybe leave out v0 to avoid confusion?

Baltram:
Thanks, and good point. I've reordered the entries and omitted the file backup version.

grote (approved):
Thanks!

PR #1039, opened 2026-07-20
[A17] Enable minify for release build

althafvly:
- Enable minify for release builds
- Add missing prebuilt for latest dav4jvm


grote:
thanks a lot for the contribution!

we haven't used minify, because of the risk of runtime crashes that are hard to test and exclude. I guess you did some manual testing? However, I'd still be worried about new issues in the future.

do you know how aosp build handles minify? does it respect the same proguard confing?

althafvly:
> thanks a lot for the contribution!
> 
> we haven't used minify, because of the risk of runtime crashes that are hard to test and exclude. I guess you did some manual testing? However, I'd still be worried about new issues in the future.
> 
> do you know how aosp build handles minify? does it respect the same proguard confing?

I only tested it in studio builds (for a few weeks now). That's why i didn't add proguards in android.bp. I can try it out soon.

grote:
Note that I am not aware of any ROM using the gradle builds. They typically built it as part of AOSP. So enabling proguard only for gradle won't help much.

You have a commit for Ktor here as well. Isolating this into its own MR would make sense. Did you notice runtime crashes without it? Or why did you need to add it?

althafvly:
> Note that I am not aware of any ROM using the gradle builds. They typically built it as part of AOSP. So enabling proguard only for gradle won't help much.
> 
> You have a commit for Ktor here as well. Isolating this into its own MR would make sense. Did you notice runtime crashes without it? Or why did you need to add it?

It shouldn't crash, also the app still uses okhttp, not using ktor changes from webdav yet. Probably needed for future if migrating. I have added changes for inline build and in initial tests it worked just fine. ( I couldn't test a17. Only tested a16)

grote:
ktor also uses okhttp, but yeah we don't  use the new ktor abstraction yet. so that sounds like the ktor commit isn't needed, right? the a17 branch should run on a16 as well.

althafvly:
> ktor also uses okhttp, but yeah we don't use the new ktor abstraction yet. so that sounds like the ktor commit isn't needed, right? the a17 branch should run on a16 as well.

For a17 studio build ktor is required. Since new dav4jvm added it. It's not there in a16 branch

I wasn't able to use a17 branch in a16. Was getting kotlin version related build error.

mikeNG:
For reference when I tried https://review.lineageos.org/c/LineageOS/android_packages_apps_Seedvault/+/467594 it decreased Seedvault apk size from 22MB to 5.6MB which is a huge amount of space when you combine it with the same change in all other system apps (which has mostly already happened)

PR #1042, opened 2026-07-21
fix(webdav): skip Nextcloud trashbin for backup cleanup

### LINKED ISSUE #1037: Use 'X-NC-Skip-Trashbin' header to delete files without trashbin when cleaning up Nextcloud WebDAV backups
Prefacing with the fact that backing up my phone to my Nextcloud instance with WebDAV has been sturdy and solid for a while, I have the annoyance that Seedvault sends a lot of files to the trash bin when cleaning up backups. This clogs up the trash bin with files that aren't in my retention policy anymore, making it difficult to browse the bin normally. I don't want to completely disable trash bin retention, I still want that option for my user files.

Nextcloud now supports WebDAV DELETEs without trashbin with a `X-NC-Skip-Trashbin` header.
- [Issue](https://github.com/nextcloud/server/issues/49605)
- [PR](https://github.com/nextcloud/server/pull/50353)

It would be awesome to have Seedvault pass this custom header on DELETEs, at least as an opt-in. An expert setting for "custom WebDAV headers" would also work.

arpitagarwal1301:
Closes #1037

## Summary

- Send `X-NC-Skip-Trashbin: true` with WebDAV cleanup DELETE requests.
- Cover both individual and whole-backup deletion paths with a focused unit test.

## Validation

- `:core:testDebugUnitTest --tests '*WebDavBackendTest*'`
- `:core:check :core:assemble :core:ktlintCheck`

### DIFF
--- a/core/src/test/java/org/calyxos/seedvault/core/backends/webdav/WebDavBackendTest.kt
+++ b/core/src/test/java/org/calyxos/seedvault/core/backends/webdav/WebDavBackendTest.kt
@@ -5,53 +5,13 @@
 
 package org.calyxos.seedvault.core.backends.webdav
 
-import at.bitfire.dav4jvm.okhttp.DavCollection
-import at.bitfire.dav4jvm.okhttp.ResponseCallback
-import io.mockk.every
-import io.mockk.mockk
-import io.mockk.mockkConstructor
-import io.mockk.unmockkConstructor
 import kotlinx.coroutines.runBlocking
 import org.calyxos.seedvault.core.backends.Backend
 import org.calyxos.seedvault.core.backends.BackendTest
-import org.calyxos.seedvault.core.backends.TopLevelFolder
 import org.junit.Test
-import kotlin.test.assertEquals
 
 internal class WebDavBackendTest : BackendTest() {
-    override val backend: Backend by lazy {
-        WebDavBackend(WebDavTestConfig.getConfig(), ".SeedvaultTest")
-    }
-
-    @Test
-    fun `delete requests skip the Nextcloud trashbin`(): Unit = runBlocking {
-        val headers = mutableListOf<Map<String, String>>()
-        mockkConstructor(DavCollection::class)
-        every {
-            anyConstructed<DavCollection>().delete(
-                ifETag = null,
-                ifScheduleTag = null,
-                headers = capture(headers),
-                callback = any(),
-            )
-        } answers {
-            arg<ResponseCallback>(3).onResponse(mockk(relaxed = true))
-        }
-
-        try {
-            val backend = WebDavBackend(
-                WebDavConfig("https://example.com", "user", "password"),
-                ".SeedvaultTest",
-            )
-            backend.remove(TopLevelFolder("folder"))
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/calyxos/seedvault/core/backends/webdav/WebDavBackendTest.kt
```diff
@@ -5,13 +5,53 @@
...
+            )
+            backend.remove(TopLevelFolder("folder"))
+            backend.removeAll()
+        } finally {
+            unmockkConstructor(DavCollection::class)
+        }
+
+        val expected = mapOf("X-NC-Skip-Trashbin" to "true")
+        assertEquals(listOf(expected, expected), headers)
+    }
```
grote: This seems to mix the integration tests with a unit test. Best to make a new test class for it.

grote (changes requested):
Thanks for the contribution. Did you actually test this against a supporting Nextcloud server?

arpitagarwal1301:
I did not test this against a live Nextcloud instance because I do not currently have a supporting server available. I verified both cleanup paths pass `X-NC-Skip-Trashbin: true` to dav4jvm with the focused unit test, and I split that test into `WebDavBackendUnitTest` as requested. The existing credential-dependent WebDAV integration tests remain unchanged and skipped locally because the `NEXTCLOUD_*` variables are unavailable. If live-server verification is required, I will need a private test instance or a locally provisioned Nextcloud environment.

tbrandirali:
@arpitagarwal1301 thank you so much for working on the feature I requested! I don't really have the time to build and test this branch myself right now, but I could probably spin up a test instance of Nextcloud for you to test it against. It might take me a few days to find the time, but if you don't have another option I'd be happy to help out.

arpitagarwal1301:
Thank you—that would be very helpful. Please don’t post credentials here. If you can prepare a temporary Nextcloud 31+ instance with a dedicated, empty test account, let me know and we can agree on a private handoff method. I’ll use it only to test this branch and confirm when the account can be deleted.

grote:
@tbrandirali this would be very appreciated! I'd prefer merging this only after it has been actually verified in practice against a real server.

tbrandirali:
After a bit more thought I think I'd rather run the test myself, spinning up a Nextcloud instance for someone else would be more work. @grote would running the integration tests against my live Nextcloud instance on a fresh account be sufficient for you, or would you require a full emulator backup test?

The integration tests do run DELETEs, so I can check that there are no leftovers in the trash bin after running them.

grote:
@tbrandirali since seedvault needs to be a system app, getting a dev setup isn't trivial. I think creating a test account on your nextcloud instance should be way easier, but as you prefer. ideally, you run the integration tests against the server and also do a manual test.

tbrandirali:
@arpitagarwal1301 I spun up a test instance for this and emailed you about exchanging credentials. @grote let me know if you would also want a test account.

althafvly:
I can confirm that it works. I tested it myself with a Nextcloud instance (WebDav).

PR #1043, opened 2026-07-23
Exclude SMS provider from backup on non-main users

althafvly:
Exclude SMS provider from backup on non-main users

https://github.com/LineageOS/android_packages_providers_TelephonyProvider/blob/lineage-23.2/AndroidManifest.xml#L73

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/transport/backup/PackageService.kt
@@ -20,12 +20,14 @@ import android.content.pm.PackageManager.GET_SIGNING_CERTIFICATES
 import android.content.pm.PackageManager.MATCH_SYSTEM_ONLY
 import android.content.pm.ResolveInfo
 import android.os.RemoteException
+import android.os.UserHandle
 import android.util.Log
 import android.util.Log.INFO
 import androidx.annotation.WorkerThread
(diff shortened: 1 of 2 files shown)


grote:
Android 16 is not really maintained anymore, can we target android17 branch with this?

cc @mikeNG since this involves system backup

althafvly:
> Android 16 is not really maintained anymore, can we target android17 branch with this?
> 
> cc @mikeNG since this involves system backup

done

grote:
Thanks! Note that your change is just hiding it from the UI. The change doesn't necessarily affect the backup of the provider nor the restore.

althafvly:
> Thanks! Note that your change is just hiding it from the UI. The change doesn't necessarily affect the backup of the provider nor the restore.

It should be good now.

mikeNG:
> Android 16 is not really maintained anymore, can we target android17 branch with this?

Well, neither LineageOS nor CalyxOS have shipped android 17 builds yet, so IMO 16 is still pretty relevant and is the branch that would be the most affected/tested by such changes.

grote:
typically, we have targeted the latest branch and then backported. if we start merging in various branches, we will end up losing commits, because we lose the overview what went in where.

mikeNG (approved):

althafvly:
@grote can we merge?

grote:
@althafvly did you test this change on any device already?

althafvly:
> @althafvly did you test this change on any device already?

Yes

PR #1044, opened 2026-07-24
Warn when Backup now has nothing selected to back up

### LINKED ISSUE #1041: Improve UX when both, app and file backup are disabled
Source of this issue: https://github.com/seedvault-app/seedvault/discussions/1033

## Desired solution

In case when user attempts to create first backup, without selected any options (no things to backup) the app should either notify about situation `There is nothing to backup`/`No resources to backup selected` or it should be not possible in the UI to enable and schedule backups.

## Current state

Right now after triggering backup nothing happens. This may be misleading for users (for example user may just enable scheduled backups and just expect it to work with defaults).

Amagah:
Fixes #1041.

## What's happening today
If both "Backup my apps" and "Backup my files" are turned off, pressing "Backup now" silently does nothing — no error, no toast, no indication that anything is wrong. That's what was reported in the linked discussion (#1033): a user configured a storage location but hadn't enabled either backup type, pressed "Backup now", and got no feedback at all.

## Fix
Before triggering a backup, check whether at least one of `backup` (app backup) or `backupStorage` (file backup) is enabled. If neither is on, show a dialog explaining that nothing is selected, instead of silently doing nothing.

To avoid duplicating the dialog-building code, I pulled the existing "metered network" warning dialog (which already covers a similar "backup didn't run, here's why" case) out into a small `showBackupInfoDialog()` helper and reused it for both cases.

## Scope
This covers the manual "Backup now" path, which is the symptom described in the issue. The issue also mentions a user enabling *scheduled* backups and expecting them to work with defaults — that path isn't touched here. Happy to follow up on it separately if you'd like it in the same change.

## Testing
- `./gradlew :app:compileDebugKotlin` — passes
- `./gradlew :app:processDebugResources` — passes (new strings valid)
- `./gradlew :app:ktlintCheck` — passes
- Manually reasoned through both branches (nothing selected vs. metered network) since I don't have a device handy to run this on right now — happy to add a screenshot if that's wanted before merge.

### DIFF
--- a/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsFragment.kt
+++ b/app/src/main/java/com/stevesoltys/seedvault/settings/SettingsFragment.kt
@@ -15,6 +15,7 @@ import android.view.MenuItem
 import android.view.View
 import android.widget.Toast
 import android.widget.Toast.LENGTH_LONG
+import androidx.annotation.StringRes
 import androidx.appcompat.widget.Toolbar
 import androidx.lifecycle.lifecycleScope
 import androidx.preference.Preference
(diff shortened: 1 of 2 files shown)


grote (approved):
great contribution. thanks!

